PR #9, opened 2017-11-27
working endpoints crm

stivni:


src/02-crm/companies.apib
```diff
@@ -0,0 +1,179 @@
...
+            + business_type: `NV` (string)
+            + vat_number: `BE0899623035` (string)
+            + local_business_number: `63326426` (string)
+            + email: `info@piedpiper.eu` (string)
+            + invoice_email: `invoicing@example.com` (string, optional)
+            + telephones (array[Telephone])
+            + website: `https://piedpiper.com` (string)
+            + avatar (string)
+            + address (Address)
+            + delivery_address (Addressee)
```
jenssegers: Missing

src/02-crm/contacts.apib
```diff
@@ -0,0 +1,208 @@
...
+Add a new contact.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + first_name: `John` (string, required)
+        + last_name: `Smith` (string, required)
+        + email: `john@example.com` (string, required)
+        + salutation: `Mr` (string, optional)
+        + email: `john@example.com` (string, required)
```
jenssegers: double `email`

src/02-crm/contacts.apib
```diff
@@ -0,0 +1,208 @@
...
+        + telephones (array[Telephone])
+        + website: `http://example.com` (string, optional)
+        + address (Address, optional)
+        + delivery_address (Addressee, optional)
```
jenssegers: Should we include `avatar`?

mrnnn: It's easy and pretty much free, right? Let's do it, at least in the info?

src/02-crm/companies.apib
mrnnn: They are there now, but are we changing this to an array before releasing? or not?

marijnvandevoorde (approved):
LGTM and iterate? Only blocking thing is the addresses

src/02-crm/companies.apib
jenssegers: I would prefer to release with an array.

stivni:
@jenssegers made another PR

PR #17, opened 2017-11-30
Add requirements and tax_id to invoices draft

luisabritotl:


apiary.apib
```diff
@@ -524,8 +524,7 @@ Draft a new invoice
                         + quantity: `3` (number, required)
                         + description: `An awesome product` (string, optional)
                         + unit_price (WritePrice, optional)
-                        + tax (object)
-                            + rate: `0.21` (number)
+                        + tax_id: `c0c03f1e-77e3-402c-a713-30ea1c585823` (string, required)
```
stivni: when the invoice is booked, this will not be enough... cause the taxRate might cease to exist... it's something we have to keep in mind...

@steftrenson @jdrieghe any thoughts on this?

jdrieghe: I don't think we should allow booking of invoices with a tax rate that is not available.
The book should be a big validation step to ensure everything is correct, no?

PR #18, opened 2017-12-01
Add discounts to line items

jdrieghe:
### Added
- discounts on invoice line items


stivni:
@jdrieghe we're not going to include invoice level discounts in the proposal?

jdrieghe:
@stivni step by step

PR #19, opened 2017-12-01
Proposal for invoice-wide discounts

steftrenson:


apiary.apib
```diff
@@ -466,6 +466,13 @@ Get a single invoice.
                     + contact (object)
                         + type: `contact` (string)
                         + id: `f88ab83c-b8ca-497b-a2c6-1ad631fe1cea` (string)
+            + discounts (array)
+                + (object)
+                    + description: `Discount` (string)
+                    + type: `percentage` (enum[string])
+                        + Members
+                            + percentage
+                    + value: `15.00` (number)
```
jenssegers: Numeric values don't need the quotes. I'm actually not quite sure when you *should* use them. My guess would be that it is only relevant to string values.

steftrenson: I think it makes more sense without them as well, but I wanted to be consistent with the other values. If you are sure they'r not needed, i'll gladly change them all

PR #23, opened 2017-12-04
Return list of emails and addresses

jenssegers:


WouterSioen:
@jenssegers can this one be merged? We're already implementing it.

jdrieghe:
Only after the merge conflicts are fixed ;)

WouterSioen:
@svenlombaert @jenssegers @jdrieghe I fixed the merge conflicts so this can be merged.

This should get merge asap imo, since it's already implemented, so the documentation is lagging behind.

PR #27, opened 2017-12-06
Proposal for taxRates.list ordering and paging

steftrenson:


jdrieghe (approved):
Looks good to me

mrtus (approved):

apiary.apib
```diff
@@ -1028,6 +1028,28 @@ Get a list of available tax rates.
 + Attributes (object)
     + filter (object, optional)
         + department_id: `080aac72-ff1a-4627-bfe3-146b6eee979c` (string, optional) - The ID of the department you wish to filter on.
+    + page (Page, optional)
+    + sort (array, optional)
+        + (object)
+            + field (enum[string], required)
+                + Members
+                    + deparment_id
```
stivni: it is technically not possible to choose from this list... so not sure if we should document it like this...

PR #32, opened 2017-12-11
Add for_attention_of to invoices update

luisabritotl:


stivni:
shouldn't it be part of the `.draft` too, before we add it to the `.update`?

jdrieghe:
It's already supported by the .draft: https://github.com/teamleadercrm/core/blob/master/src/Invoicing/Domain/Invoices/DraftInvoice.php#L76
We just need to add it to the docs still. Can you make a PR for this @luisabritotl ?

WouterSioen:
@luisabritotl I think this is already implemented? Shouldn't this get merged? (and conflicts fixed though)

luisabritotl:
Done

PR #34, opened 2017-12-13
webhooks.register

wouterds:
Tested with specs & already live, behaves as documented.


mrtus (dismissed):

WouterSioen:
@stivni We changed the accepted webhook types here to mimic the new API structure but couldn't find anything about these webhooks:

* TASK_FINISHED
* FILE_ADDED
* FILE_EDITED
* FILE_DELETED
* USER_DAYS_OFF_EDITED
* COMPANY_CLOSING_DAYS_EDITED

What should we do with those? Just convert them to something that looks like the API's endpoints? (f.e. `task.finished`, `file.added`,...) even though they will probably not be called by the new API (yet)?

Jdww:
I think this endpoint is already used so if we change this some people should probably be notified. @nielsvandenbuverie knows more about this.

WouterSioen:
@Jdww we would still support the old types as well, but undocumented.

WouterSioen:
Poke @stivni

wouterds:
Poke @jenssegers

WouterSioen:
Picked this one up again since we're going to build this now.

@jenssegers @stivni would like your 👍 before we start implementing it.

We will probably have to add more webhook types later, but we can start with the ones that are support from the new API and always add more later I guess?

jenssegers (dismissed):

stivni:
I would even say to exclude even more webhooks and to add them gradually (by every separate team), but in general, this looks ok

WouterSioen:
@stivni I think we can easily support all those events from the start since we already support them in the service anyway.

src/99-other/webhooks.apib
```diff
@@ -0,0 +1,45 @@
+## Webhooks [/webhooks]
+
+### webhooks.register [POST /webhooks.register]
+
+Register a new webhook.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + url: `https://www.mywebhookurl.com` (string, required) - Your webhook url
```
dietrichm: PP: use `example.com` or any of the other reserved domains for example URLs (and "URL" with caps 👼).

 :ok: by @WouterSioen on 21/06/2018 13:02:06

src/apiary.apib
```diff
@@ -74,5 +74,6 @@ HOST: https://api.teamleader.eu
 # Group Other
 
 :[Migrating](./99-other/migrating.apib)
+:[Migrating](./99-other/webhooks.apib)
```
dietrichm: PP: add a blank line first, since this now adds the header directly on a line after another request's response data in `apiary.apib`. 

 :ok: by @WouterSioen on 21/06/2018 13:02:09

dietrichm (dismissed):

dietrichm (dismissed):

wouterds:
IMG
😂😄

rathesDot (dismissed):

JordyDeGroeve (approved):

dietrichm (approved):

rathesDot (approved):

PR #35, opened 2017-12-14
Add grouped lines to invoice update

luisabritotl:


WouterSioen:
@jdrieghe @steftrenson Looks like this is already implemented in the codebase, so we should probably merge this?

PR #46, opened 2018-01-11
Add payment term to invoices info

luisaterrabrito:


apiary.apib
```diff
@@ -1110,6 +1110,13 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+            + payment_term (object, required)
+                + type (enum[string])
+                    + Members
+                        + cash - Direct payment, often cash
+                        + end_of_month - End of the month of Xth day after the invoice date
+                        + after_invoice_date - X days after invoice date
+                + days (number, optional) - Modifier "X" for the above statements. Not required when type is 'cash'.
```
WouterSioen: Q: Isn't it possible to make payment_term a data structure so it doesn't have to be duplicated?

 :ok: by @luisabritotl on 18/01/2018 10:13:09

apiary.apib
```diff
@@ -1110,6 +1110,7 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+            + payment_term (PaymentTerm, required)
```
jenssegers: Not `required`

PR #47, opened 2018-01-16
Add a proposal for the tags.list endpoint

WouterSioen:


apiary.apib
```diff
@@ -825,6 +825,22 @@ Get a quotation
                 + type: `deal` (string)
                 + id: `53474a7a-f9b2-4dd4-88a8-40ce773c7a64` (string)
 
+## Tags [/tags]
+
+### tags.list [GET /tags.list]
+
+Get a list of tags.
+
++ Request (application/json;charset=utf-8)
```
stivni, line 834: B : because we can never predict amounts and order in list calls, we have a _guideline_ to always provide paging and a default sorting method on list calls... can you add this?

 :ok: by @WouterSioen on 17/01/2018 13:01:50

apiary.apib
```diff
@@ -825,6 +825,22 @@ Get a quotation
...
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `ee6fe9fa-bf92-0d25-9d10-f4e34ba1f7d1` (string)
```
stivni: Q : I also though we were going to address tags as values, not as entities... or is something blocking that? maybe take this discussion offline?

WouterSioen: @stivni I had no idea how tags would be tackled but got the task to make the `tags.list` endpoint, so I made a proposal that seemed most logical to me (because they are handled as entities in the database and can be coupled to multiple items).

I don't mind changing this though. It's just a small change.

 :ok: by @WouterSioen on 17/01/2018 13:01:51

WouterSioen:
@stivni I handled the remarks. Any more input on this?

src/02-crm/tags.apib
```diff
@@ -0,0 +1,29 @@
...
+                + (object)
+                    + field: `name`
+                    + order: `asc`
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + name: `Client` (string)
```
jenssegers: Can we lowercase all tags?

jenssegers:
Current proposal:

```json
{
	"data": [
		{
			"name": "newsletter"
		},
		{
			"name": "facebook"
		}
	]
}
```

src/02-crm/tags.apib
WouterSioen: They aren't lowercased in the UI, so would this mean that when we move to using the new API internally, all tags will have to be lowercased? Not sure if this is a behavioural change we want?

jenssegers: We want tags to become unique as well, I prefer case-insensitive tags.

 :ok: by @WouterSioen on 18/01/2018 10:33:55

WouterSioen:
This should probably only be merged after the end-to-end tests and the implementation are merged.

src/02-crm/tags.apib
```diff
@@ -0,0 +1,29 @@
...
+                + (object)
+                    + field: `tag`
+                    + order: `asc`
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + tag: `campaign` (string)
```
mrtus, line 29: I think we wanted an array of `string`s before?

But I think this allows us to add extra fields later. I do however have a remark on the key name. I think we should consider the object being the `Tag` that has a `name`.

Otherwise if we do not want to have extra fields for a `Tag` we should probably just return an array of strings?

WouterSioen: We do it like this so we can easily add extra keys. I personally prefer name as key over tag, but Jens prefers it the other way around :D

mrtus: What I was expecting yes 😄

Maybe the key name is something to rediscuss?

WouterSioen: It doesn't really matter that much, does it?

mrtus: Depends on if we want to have a `Tag` model containg a field `tag`

WouterSioen: poke @stivni @jenssegers @jdrieghe (not sure who else helps take decisions about the api, I guess all opinions are welcome)

marijnvandevoorde (approved):
LGTM

wouterds (approved):

mrtus (approved):

PR #49, opened 2018-01-22
Allow partial updates on companies.update

jenssegers:


WouterSioen (approved):
@jenssegers This should probably get merged and deployed since these changes are merged in core and will be included in the next release?

PR #50, opened 2018-01-23
Add grouped lines to quotations info

luisaterrabrito:


apiary.apib
```diff
@@ -824,6 +824,29 @@ Get a quotation
             + deal (object)
                 + type: `deal` (string)
                 + id: `53474a7a-f9b2-4dd4-88a8-40ce773c7a64` (string)
+            + grouped_lines (array)
+                + (object)
+                    + section (object)
+                        + title (string)
+                    + line_items (array)
+                        + (object)
+                            + product (object, nullable)
```
jenssegers, line 833: Not sure if we should add `nullable` to read endpoints. But will do followup pr if we figure this out.

jenssegers (approved):

apiary.apib
WouterSioen: I don't think `nullable` and `optional` are displayed in the documentation anyway?

WouterSioen (approved):

PR #55, opened 2018-01-26
Update invoices.info and add registerPayment

luisaterrabrito:
Add payments and total_due to invoices.info

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1142,7 +1142,13 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+                + total_due: 0 (number)
             + payment_term (PaymentTerm)
+            + payments (array)
+                + (object)
+                    + paid_at: `2016-03-03T16:44:33+00:00` (string)
+                    + payment (object)
+                        + Include Money
 
 ### invoices.draft [POST /invoices.draft]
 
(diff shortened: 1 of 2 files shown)


WouterSioen (approved):

mrtus (approved):

apiary.apib
```diff
@@ -1142,7 +1142,13 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+                + total_due: 0 (number)
```
jenssegers: Not a big fan of this naming. Would rather have this just `due`.

apiary.apib
```diff
@@ -1256,6 +1262,21 @@ Delete an existing invoice. Only possible for draft invoices or the last booked
...
 + Response 204
 
+### invoices.registerPayment [POST /invoices.registerPayment]
+
+Register a payment for an invoice.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `7abb325c-e063-42a4-8fb4-1b730759645a` (string, required)
```
jenssegers, line 1272: I thought we discussed we would register payments using a reference rather than introducing payment uuids?

jenssegers (dismissed):
Can you check my remarks?

apiary.apib
luisaterrabrito: will change this

 :ok: by @luisabritotl on 31/01/2018 14:24:38

apiary.apib
luisaterrabrito: I thought we were going to use the uuid, will bring up the discussion in the channel


 :ok: by @luisabritotl on 31/01/2018 14:44:57

apiary.apib
WouterSioen: I guess this change will also need to be reflected in the end-to-end tests?

luisaterrabrito: done

 :ok: by @luisabritotl on 31/01/2018 14:44:49

apiary.apib
stivni: btw, this `id` is the invoice id :p

 :ok: by @stivni on 01/02/2018 17:03:54

apiary.apib
```diff
@@ -1142,7 +1142,13 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+                + due: 0 (number)
```
stivni, line 1145: Q : doesn't `due` need to be added to the `.list` call too?

luisaterrabrito: I guess, but it isn't there yet

jenssegers (approved):

PR #56, opened 2018-01-29
Add currency to invoices draft

luisaterrabrito:


mrtus (approved):

apiary.apib
```diff
@@ -1173,6 +1173,7 @@ Draft a new invoice.
                                 + Members
                                     + excluding
                         + tax_rate_id: `c0c03f1e-77e3-402c-a713-30ea1c585823` (string, required)
+        + Include Currency
```
jenssegers, line 1176: Do we need to include this at this level?

luisaterrabrito: it is the invoice currency, so I think it makes sense to be in this level

luisaterrabrito:
Closing this until we can figure out what we want to do

PR #60, opened 2018-01-30
Add documentation for the companies.tag endpoint

WouterSioen:
Copy pasted this from the api-future repository. If approved, I'll create the end-to-end tests and implementation.

### DIFF
--- a/src/02-crm/companies.apib
+++ b/src/02-crm/companies.apib
@@ -199,3 +199,15 @@ Delete a company.
         + id: `f3d67f3e-b8a9-45e8-99b1-17a3b14de8a6` (string, required)
 
 + Response 204
+
+### companies.tag [POST /companies.tag]
+
+Add a new or existing tag to a company.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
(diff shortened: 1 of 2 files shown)


mrtus (approved):

vanbrabantf (approved):

JordyDeGroeve (approved):

jenssegers (approved):

src/02-crm/companies.apib
```diff
@@ -199,3 +199,15 @@ Delete a company.
         + id: `f3d67f3e-b8a9-45e8-99b1-17a3b14de8a6` (string, required)
 
 + Response 204
+
+### companies.tag [POST /companies.tag]
+
+Add a new or existing tag to a company.
```
jenssegers, line 205: Can you double check with the implementation if this is correct? From what I remember we would release without the ability to create new tags first?

jenssegers (changes requested):

src/02-crm/companies.apib
WouterSioen: Already implemented it **with** adding new tags, so I guess this can be merged as is (when the implementation is released)

jenssegers (approved):

PR #61, opened 2018-02-01
Add optional date to invoices draft and update

luisaterrabrito:


apiary.apib
```diff
@@ -1182,6 +1182,7 @@ Draft a new invoice.
                                 + Members
                                     + excluding
                         + tax_rate_id: `c0c03f1e-77e3-402c-a713-30ea1c585823` (string, required)
+        + date: `2016-02-04` (string, optional)
```
jenssegers: I would prefer naming this `invoice_date` so that there's no confusion what this option means.

jenssegers (dismissed):

apiary.apib
jdrieghe: I remember having this discussion a couple of times already with @stivni for other invoicing endpoints. Each time the conclusion was: let's just use `date` because we wouldn't use `invoice_id`, `invoice_total`, `invoice_status` etc.
Do you suggest we change it in the result of the `/invoices.info` and `/invoices.list` call too?
If so, can we write it down somewhere so we don't have the discussion again in a couple of weeks?

stivni: in the domain the prefix "invoice_" is important for "date" & "number"... so please stick with "invoice_date" in this case only (domain > consistency)

jdrieghe: Alright, my bad. Faulty memory it seems. Checked other invoicing endpoints and we use `invoice_date` indeed. @luisabritotl can you change this here and in the implementations?

stivni: (fyi, the business talks about _invoice number_ and _invoice date_ and not of _the number of the invoice_ like _the number of the project_ which are two different things)

Although I'm not 100% convinced this is the best approach, I am sure it suits better than _date_... `issue_date` would suit better for booked invoices, but not for drafts, so ... difficult...

 :ok: by @luisabritotl on 05/02/2018 14:30:22

mrtus (approved):

jdrieghe (approved):

jdrieghe:
Can be merged after https://github.com/teamleadercrm/core/pull/4146 is deployed 👍

PR #72, opened 2018-02-21
[CORE-442] [CORE-752] Add companies.list: filter on tags

dietrichm:
Add docs for the added filter on tag in companies.list.

❗️ only merge when [CORE-752] is merged.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -611,7 +611,7 @@ Get a list of companies.
                 + email: `info@piedpiper.eu` (string, required)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `Acme` (string, optional) - Filters on name, vat number, emails and telephones
-            + tag: `expo` (string, optional) - Filters on tag name
+            + tags: `expo`,`lead` (array[string], optional) - Filters on tag name
         + page (Page, optional)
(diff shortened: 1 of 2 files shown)


WouterSioen (approved):
👍 But let's wait to merge 'till the implementation is deployed.

JordyDeGroeve (approved):

apiary.apib
```diff
@@ -611,6 +611,7 @@ Get a list of companies.
                 + email: `info@piedpiper.eu` (string, required)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `Acme` (string, optional) - Filters on name, vat number, emails and telephones
+            + tag: `expo` (string, optional) - Filters on tag name
```
jenssegers: Can we change this to `tags`? So that we can search on multiple tags at the same time? Kinda makes more sense to me :)

jenssegers (changes requested):

apiary.apib
dietrichm: We will make a follow up PR in core to handle this.
The docs are updated.

 :ok: by @WouterSioen on 22/02/2018 11:05:18

WouterSioen (approved):

WouterSioen:
@jenssegers can this be unblocked and merged? This is released

jenssegers (approved):

PR #76, opened 2018-03-01
Add timetracking endpoints proposal

jenssegers:


apiary.apib
```diff
@@ -1855,6 +1855,272 @@ Update a milestone.
...
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + ended_at: `2017-04-26T11:01:49+00:00` (string, required)
+            + Properties
+                + started_on: `2017-04-26` (string, required)
+                + duration: 3600 (number, required) - In seconds
+            + Properties
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + duration: 3600 (number, required) - In seconds
+        + description (string, optional)
+        + subject (object, optional)
```
WouterSioen, line 2769: Q: shouldn't all these properties be nullable to allow partial updates?

 :ok: by @WouterSioen on 19/03/2018 09:58:04

WouterSioen (dismissed):
Let's do this

apiary.apib
```diff
@@ -2192,6 +2192,272 @@ Update a milestone.
...
+                    + milestone
+                    + ticket
+            + id: `29ff471c-7d8f-40d5-8c95-9a9cab841e65` (string, required)
+
++ Response 201
+
+    + Attributes (object)
+        + data (object)
+            + type: `timetracking` (string)
+            + id: `4ff475b2-f45b-4803-800e-350b9c02aeca` (string)
```
WouterSioen: Is it just me or is it strange that we return the same type here for an "ongoing timetracking" as we do for a "submitted timetracking", even though we can't fetch an ongoing timetracking using timetracking.info?

It would be more logical to me to name it ongoingTimeTracking as long as it's not submitted and thus fetchable in timeTracking.info.

apiary.apib
```diff
@@ -2192,6 +2192,272 @@ Update a milestone.
...
+
+### timetracking.stop [POST /timetracking.stop]
+
+Stop the current timetracking timer.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + ended_at: `2017-04-26T10:01:49+00:00` (string, optional) - If not provided, current time will be used
```
WouterSioen: Shouldn't this endpoint return the `id` of the added timetracking?

This will be different then the `id` returned from `timetracking.start`, because it's another datatype (ongoing vs finished timetracking) + the stopped timetracking can be split here when it spanned multiple days.

 :ok: by @WouterSioen on 28/03/2018 10:30:11

apiary.apib
jenssegers: On the other hand, of we don't return an id, you are also not able to update information related to a running timer. For me it's ok to change the proposal to return a `204 No Content`, but then we need to be aware of that too.

WouterSioen: So you meant the `timetracking.info` and `imetracking.update` where meant to also work with "ongoing" timetrackings?

I can see the use case in allowing it for `timetracking.update`, because otherwise you can't edit f.e. the description of a running timer.

 :ok: by @WouterSioen on 28/03/2018 10:30:05

WouterSioen:
Up for re-approve @jenssegers @stivni @mrnnn @JordyDeGroeve @dietrichm 

Note that I didn't add `timer.update` yet in this PR since this can be released later in a separate PR as an extra feature.

JordyDeGroeve (approved):
LGTM. Renaming to `Timer` makes things clearer.

apiary.apib
```diff
@@ -2192,6 +2192,277 @@ Update a milestone.
 
 + Response 204 (application/json;charset=utf-8)
 
+# Group Timetracking
```
jenssegers: Can we make this `Time Tracking`? Same for title below.

dietrichm (approved):

JordyDeGroeve (dismissed):

dietrichm (dismissed):

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,268 @@
...
+Get a list of tracked time.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + ids: `404c1777-6b61-4368-a388-0301dde8e72a`,`409e3446-96f2-4d9a-b20b-7f7cd055053e` (array[string], optional)
+            + user_id: `87982c96-f2fe-4b05-838c-ff42c0525758` (string, optional)
+            + work_type_id: `ce1287f0-0480-458d-b371-22862d864074` (string, optional)
+            + tracked_after: `2017-04-26T10:01:49+00:00` (string, optional)
```
JordyDeGroeve: Q: We filter on `started_after` in the code. Should we update this?

https://github.com/teamleadercrm/service-time-tracking/pull/60

WouterSioen: Yep, you're right. That's what the Jira ticket said and I didn't double check with the documentation.

I think `started_after` is a bit more explicit though.

src/06-time-tracking/time-tracking.apib
```diff
@@ -176,6 +176,23 @@ Delete tracked time.
 
 + Response 204
 
+### timeTracking.resume [POST /timers.resume]
```
dietrichm: B: should be `POST /timeTracking.resume` I think?

dietrichm (changes requested):

apiary.apib
```diff
@@ -2192,6 +2192,277 @@ Update a milestone.
...
++ Request (application/json)
+
+    + Attributes (object)
+        + ended_at: `2017-04-26T10:01:49+00:00` (string, optional) - If not provided, current time will be used
+
++ Response 201 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + type: `timeTracking` (string)
```
mansilaxel, line 2801: Q: type: `timer` (string)? It is like this in `timers.start`

src/06-time-tracking/time-tracking.apib
WouterSioen: true that

 :ok: by @WouterSioen on 17/04/2018 16:41:40

apiary.apib
WouterSioen: Nope, when you stop a timer, it get's saved, which makes it become a timeTracking

 :ok: by @WouterSioen on 17/04/2018 16:42:09

JordyDeGroeve (dismissed):

dietrichm (dismissed):

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,268 @@
...
+        + started_at: `2017-04-26T10:01:49+00:00` (string, optional) - If not provided, current time will be used
+
++ Response 201
+
+    + Attributes (object)
+        + data (object)
+            + type: `timer` (string)
+            + id: `f01adf4a-bb9b-45de-b231-615cd0e941de` (string)
+
+### timers.current [GET /timers.current]
```
dietrichm: PP: A header for "Timers" should be added before this line.

 :ok: by @dietrichm on 25/04/2018 17:19:10

dietrichm (dismissed):

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,270 @@
...
+
+    + Attributes (object)
+        + work_type_id: `2175597d-484e-4a1c-a781-cbc3d9f893ba` (string, optional)
+        + One Of
+            + Properties
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + ended_at: `2017-04-26T11:01:49+00:00` (string, required)
+            + Properties
+                + started_on: `2017-04-26` (string, required)
+                + duration: 3600 (number, required) - In seconds
```
marijnvandevoorde: B: This has to be removed from docs for now!!!

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,270 @@
...
+Add tracked time.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + work_type_id: `2175597d-484e-4a1c-a781-cbc3d9f893ba` (string, optional)
+        + One Of
+            + Properties
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + ended_at: `2017-04-26T11:01:49+00:00` (string, required)
```
marijnvandevoorde: B: not in there yet, but can be delivered quite fast I think, so let's keep this in here. But blocked until implemented

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,270 @@
...
+
+    + Attributes (object)
+        + id: `66621f54-3d0d-420f-8d4b-ddd7fc473bf2` (string, required)
+        + work_type_id: `a12f4289-3580-4780-bb6d-b4e9f84f80ad` (string, optional, nullable)
+        + One Of
+            + Properties
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + ended_at: `2017-04-26T11:01:49+00:00` (string, required)
+            + Properties
+                + started_on: `2017-04-26` (string, required)
```
marijnvandevoorde: B: remove this

src/06-time-tracking/time-tracking.apib
```diff
@@ -0,0 +1,270 @@
...
+Update tracked time.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `66621f54-3d0d-420f-8d4b-ddd7fc473bf2` (string, required)
+        + work_type_id: `a12f4289-3580-4780-bb6d-b4e9f84f80ad` (string, optional, nullable)
+        + One Of
+            + Properties
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
```
marijnvandevoorde, line 137: B: until implemented. But let's keep this for now.

dietrichm (dismissed):

marijnvandevoorde (changes requested):
@jenssegers Can we remove the 2nd way of adding tracking (fuzzy, no started_at, only started_on). We don't have it yet.

JordyDeGroeve:
Can we remove the `.delete` endpoint until the side effects are handled?
If we want to release this asap.

WouterSioen:
Updated to be compatible with our current implementation, should be good to merge now.

WouterSioen (approved):

JordyDeGroeve (approved):
👌

mansilaxel (approved):

mansilaxel:
Not sure about using `tracked time` though.

src/06-time-tracking/time-tracking.apib
```diff
@@ -100,12 +100,6 @@ Add tracked time.
     + Attributes (object)
         + work_type_id: `2175597d-484e-4a1c-a781-cbc3d9f893ba` (string, optional)
         + One Of
-            + Properties
```
marijnvandevoorde: This makes no sense, right? let's remove this level and require started_at & duration for now

marijnvandevoorde (dismissed):

src/06-time-tracking/time-tracking.apib
marijnvandevoorde: agreed to do this if needed, or add the ended_at soon so this is no longer an issue :-)

PR #88, opened 2018-03-13
Add documentation about rate limiting

jenssegers:


apiary.apib
```diff
@@ -80,6 +80,14 @@ Errors are always returned as an array of error objects under an `errors` top le
 }
 ```
 
+### Rate limiting
+
+To be able to ensure a predictably pleasant experience for our users, all API endpoints are rate limited. API calls are evaluated per integration (client id), with a maximum of **100 calls** in a window of **1 minute**.
```
jdrieghe: This section is rather bloated. Note that this is technical documentation and we should make it as easy to read/scan as possible. How about something like:

> To ensure a fast and predictable experience for everyone, all our API endpoints have a rate limit of **100 calls per minute** for each integration (or client id).

jenssegers: Done!

PR #91, opened 2018-03-15
Replace dev scripts with Makefile

jenssegers:


JordyDeGroeve (approved):
nice!

Makefile
```diff
@@ -0,0 +1,18 @@
+PWD = $(shell pwd)
+
+all: lint build
+
+node_modules: package.json
+	docker run -it --init --rm -v $(PWD):/code -w /code node npm install
```
stivni, line 6: can you move this to a `bin/npm` too?

steftrenson (approved):

Makefile
jenssegers: If this was a node project I would. But since you never actually have to use `npm` in this project I would rather not introduce a `bin` folder that might be confusing. I kinda like I could remove te dev/bin folder to make the project more clean again :)

PR #92, opened 2018-03-15
Add changelog section

jenssegers:


src/apiary.apib
```diff
@@ -7,6 +7,8 @@ HOST: https://api.teamleader.eu
 
 :[OAuth2](oauth2.apib)
 
+:[Changelog](changelog.apib)
```
mrtus, line 10: Something to add to the bottom maybe?

As a developer I wouldn't want to scroll through a big changelog just to see the first endpoint...

jdrieghe (approved):

src/apiary.apib
jenssegers: I don't think we're going to embed the entire changelog within the documentation. We will probably just link to a `CHANGELOG.md` file then.

mrtus: That's something I can live with 👍

mrtus (approved):

JordyDeGroeve (approved):

steftrenson (approved):

PR #97, opened 2018-03-19
[CORE-675] Return custom_fields in companies.info

WouterSioen:


apiary.apib
```diff
@@ -778,6 +778,12 @@ Get details for a single company.
             + added_at: `2016-02-04T16:44:33+00:00` (string)
             + updated_at: `2016-02-05T16:44:33+00:00` (string)
             + tags: prospect, expo (array[string])
+            + custom_fields (array)
+                + (object)
+                    + field: (object)
+                        + type: `customField` (string)
```
jenssegers: If we go for `customFieldDefinition` endpoints, we should also name this `customFieldDefinition`. @stivni 

 :ok: by @WouterSioen on 22/03/2018 16:47:35

WouterSioen: This is how it was implemented, but just as optional include. I think it was needed for @mrtus 

I think we should also include more information here like the name of the custom field. This will the returned information a lot more logical and useful.

mrtus: Please don't break my custom fields 🙃  (I don't mind about the type though, we don't use that)

WouterSioen: @mrtus adding more properties like the name of the custom field couldn't really hurt, right?

mrtus: Offcourse not :) adding should never break an API

src/02-crm/companies.apib
```diff
@@ -112,6 +112,12 @@ Get details for a single company.
             + added_at: `2016-02-04T16:44:33+00:00` (string)
             + updated_at: `2016-02-05T16:44:33+00:00` (string)
             + tags: prospect, expo (array[string])
+            + custom_fields (array)
+                + (object)
+                    + field: (object)
+                        + type: `customField` (string)
+                        + id: `bf6765de-56eb-40ec-ad14-9096c5dc5fe1` (string)
+                    + value: `092980616` (string)
```
WouterSioen: I'm also not sure about this. If it's a custom field that links to a contact (or other "other_id"), it should probably return the type and the id inside an object? // @jenssegers @stivni @mrtus

mrtus: Custom field layout should probably depend on the type of the custom field.

WouterSioen:
I added a proposal for the enum and object custom fields in the latest commit. Discussion required.

apiary.apib
jenssegers: @WouterSioen We can't embed data inside that `customFieldDefinition`. The `type` field of the  definition would conflict with the type already used here. For relations we return a type/id "tuple" that represents the relationship, but it will never hold data.

If we want to include information about the custom fields, we should side-load it and put it at the bottom of the page. Which in the end is the same as doing a `customFieldDefinitions.list` call with a filter.

src/02-crm/companies.apib
stivni: but don't you always have to choose the type `customer_id` for example?

stivni (dismissed):
I'm not convinced by how we return the values... I think it should consistently return a value property. Let's discuss the different cases we need to support irl.

jenssegers:
I agree with @stivni on this one.

WouterSioen:
@jenssegers @stivni let's do a meeting. (maybe include @mrtus too since he's already consuming custom fields in some internal tools?)

src/02-crm/companies.apib
WouterSioen: But we don't pass the type of the custom field. I would expect to be able to see if I get a scalar value, a multiselect or an entity here, without having to query customFieldDefinition endpoints.

```
{
    "custom_fields": [
       // normal field
        {
            "field": {
                "type": "customFieldDefinition",
                "id": "bf6765de-56eb-40ec-ad14-9096c5dc5fe1"
            },
            "value": {
                "type": "string",
                "value": "092980616"
            }
        },
       // multiple select option
        {
            "field": {
                "type": "customFieldDefinition",
                "id": "bf6765de-56eb-40ec-ad14-9096c5dc5fe1"
            },
            "value": {
                "type": "multipleSelect",
                "value": "selected-value"
            }
        },
       // entity field
        {
            "field": {
                "type": "customFieldDefinition",
                "id": "bf6765de-56eb-40ec-ad14-9096c5dc5fe1"
            },
            "value": {
                "type": "contact",
                "value": "bf6765de-56eb-40ec-ad14-9096c5dc5fe1" // contact uuid
            }
        }
    ],
}
```

This might be a way to return all types in the same structure with the type information available?

mrtus: Doesn't look bad! But I am wondering if we really need the type (i'd say yes go for it, you can see it at a glance when using the api) but the value type is also already defined in the cf definition. To be discussed atleast 👍 But I do like this aproach! This would surely help integrators with parsing custom fields.

WouterSioen: The double "value" is a bit strange though. The type might be more logical in the "field" array 🤔 

 :ok: by @WouterSioen on 23/03/2018 11:45:35

apiary.apib
```diff
@@ -778,6 +778,25 @@ Get details for a single company.
...
             + tags: prospect, expo (array[string])
+            + custom_fields (array)
+                + (object)
+                    + field: (object)
+                        + type: `customFieldDefinition` (string)
+                        + id: `bf6765de-56eb-40ec-ad14-9096c5dc5fe1` (string)
+                    + One Of
+                        + value: `092980616` (string)
+                        + Properties:
+                            + type: `company` (enum[string])
```
stivni:
```
+ One Of
    + value (string) - For ...
    + value (bool) - For ...
    + value (object) - For ...
        + type (string)
        + id (string)
```

 :ok: by @WouterSioen on 23/03/2018 11:45:30

apiary.apib
```diff
@@ -778,6 +778,25 @@ Get details for a single company.
...
+                                + Members
+                                    + company
+                                    + contact
+                                    + product
+                                    + user
+                            + id: `bf6765de-56eb-40ec-ad14-9096c5dc5fe1` (string)
+                        + Properties:
```
stivni: remove

 :ok: by @WouterSioen on 26/03/2018 07:59:31

WouterSioen:
New proposal @stivni @jenssegers @JordyDeGroeve @mrtus @dietrichm

dietrichm (approved):

mrtus (approved):

JordyDeGroeve (approved):

apiary.apib
```diff
@@ -778,6 +778,23 @@ Get details for a single company.
             + added_at: `2016-02-04T16:44:33+00:00` (string)
             + updated_at: `2016-02-05T16:44:33+00:00` (string)
             + tags: prospect, expo (array[string])
+            + custom_fields (array)
+                + (object)
+                    + field: (object)
```
jenssegers: Just thinking out loud here; since we decided to name this a "definition", maybe we should also rename "field" to "definition"?

WouterSioen: Hmm, interesting idea. That would be more consistent indeed. If others also think this is a good idea, I'd change it. 👍 

 :ok: by @WouterSioen on 26/03/2018 10:35:26

mrtus (approved):

stivni (approved):

jenssegers (approved):

dietrichm (approved):

PR #102, opened 2018-03-21
Add contact_id to events.list filters

Jdww:


WouterSioen (approved):

psrpinto (approved):

mrtus (approved):

src/04-calendar/events.apib
```diff
@@ -16,6 +16,7 @@ Get a list of calendar events.
             + activity_type_id: `edd94120-63e2-4b10-80ed-fdfcfaa0a515` (string, optional)
             + ends_after: `2017-01-01T00:00:00+00:00` (string, optional) - Start of the period for which to return events
             + starts_before: `2018-01-01T00:00:00+00:00` (string, optional) - End of the period for which to return events
+            + contact_id: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5` (string, optional)
```
mrtus, line 19: Won't we need the possibility of requesting events for multiple contacts? for e.g. comparing a set of contacts to see if they are free at the same date?

WouterSioen: You can easily do this with multiple requests for now?

jenssegers (changes requested):
What does this filter apply to? The attendees or the links? Probably best to rename the filter to either `attendee` or `link`.

Jdww:
@jenssegers I discussed this with @bbouwmeister but I think it will be quite clear once we've attended the filter for attendee contacts called `attendee_contact` and we'll have more links in the future like company, contact, project, ...

jenssegers:
@Jdww I would rather see a filter like:

```json
{
	"filter": {
		"attendees": {
			"type": "contact",
			"id": "cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5"
		}
	}
}
```

This is the same format as we have for `customer` filtering on `deals.list`: https://developer.teamleader.eu/#/reference/deals/deals/deals.list

Jdww:
@jenssegers This way you wouldn't be able to filter on multiple links.
```
{
	"filter": {
		"links": {
			"contact_id": "cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5",
                        "project_id": "cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5"
		}
	}
}
```
Wouldn't something like this make more sense in that case?

jenssegers:
@Jdww If we really want to be able to filter on multiple links;

```
{
	"filter": {
		"links": [
			{
				"type": "contact",
				"id": "cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5"
			},
			{
				"type": "project",
				"id": "cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5"
			}
		]
	}
}
```

PR #107, opened 2018-03-22
Add attendees to events.list

psrpinto:


apiary.apib
```diff
@@ -961,7 +961,7 @@ Get a list of deals.
         + filter (object, optional)
             + ids: `3f507137-e599-41d4-8625-0e2adba0be4c`,`ae97f6d3-0bb4-43d3-b13d-f06069d07e98` (array[string], optional)
             + customer (object, optional)
-                + type: `contact` (enum[string], required)
+                + type (enum[string], required)
```
JordyDeGroeve, line 964: Q: Why is `contact` removed?

psrpinto: Not sure why contact was specified like that. An example maybe?

jenssegers (approved):

apiary.apib
WouterSioen: You can click "sample" under the options then, and it will show `contact` if provided.

I don't find it really useful though for enum fields. All values are "sampled" already in the list.

 :ok: by @regularjack on 26/03/2018 11:47:29

WouterSioen (approved):

JordyDeGroeve (approved):

PR #110, opened 2018-03-26
More information about departments

jenssegers:
Related Implementation PR: https://github.com/teamleadercrm/core/pull/6753
Related E2E PR: https://github.com/teamleadercrm/api-tests/pull/410

Integrators might want to get information about the company of the user to set up an initial account on their end, or for invoicing reasons.


WouterSioen (dismissed):

dietrichm (dismissed):

JordyDeGroeve (dismissed):

marijnvandevoorde:
@steftrenson, WDYT: should we add currency here as well while we're at it?

Other than that, looking fine!

steftrenson:
@mrnnn You could already do that, although I have no idea on what the timeframe is to actually have a department in a different currency

steftrenson (dismissed):

marijnvandevoorde (dismissed):
Let's implement this.

So far we have id, name & currency

thomasnoe07:
@jenssegers Can we move this further or it should stay open?

luisaterrabrito:
Why the same information in the list and info?

dcsg:
I agree with @luisabritotl, why do we have the same information in `list` and `info`? Also, why the `website`?

apiary.apib
```diff
@@ -291,7 +291,25 @@ Get a list of departments.
...
+                + address (Address)
+                + emails (array)
+                    + (object)
+                        + type: `primary` (enum[string])
+                            + Members
+                                + primary
+                                + invoicing
+                        + email: `info@piedpiper.eu` (string)
+                + telephones (array[object])
+                    + type: `phone` (enum[string], required)
```
paultimoce: This should be mobile i think since the corresponding database field is gsm

paultimoce (changes requested):

apiary.apib
WouterSioen: only mobile and fax for departments for now

jenssegers: In the UI it's called telephone and fax, hence the phone here. Phone also makes more sense on company level (imo).

WouterSioen: 😞

jenssegers:
I see no problem in having the same information returned in both the list and info endpoints. It's basically the same information we're returning here as for the contact/company endpoints.

dcsg:
@jenssegers why iban, website, bic and that type of information is available in a `list` endpoint? It belongs to the **details** of the `resource` hence in the `info` endpoint. Otherwise, why do we have two endpoints? We could just have the `list` one.

apiary.apib
```diff
@@ -308,7 +326,25 @@ Get details for a single department.
...
+                    + email: `info@piedpiper.eu` (string)
+            + telephones (array[object])
+                + type: `phone` (enum[string], required)
+                    + Members
+                        + phone
+                        + fax
+                + number: `092980615` (string, required)
+            + website: `https://piedpiper.com` (string)
             + currency: `EUR` (string)
+            + iban: `BE12123412341234` (string)
```
dcsg, line 329: B/Q: why are we adding this to the root level? This for me seems to be missing a "bank information" object and then having these fields inside that object.

dietrichm: Lots of endpoints are adding IBAN and BIC on the top level right now. This might be a BC breaking improvement that we can release in a next API version.

luisaterrabrito:
I don't agree, @jenssegers. I understand more information might be added in the future to the info endpoint, but for now, I find it confusing, I wouldn't know which one to use. Feels like there is no need to have both.

jenssegers:
Ok let's reduce the information on the list endpoint then. Easier to add it later than to remove it afterwards. Any other suggestions for fields to drop other than the iban/bic?

luisaterrabrito:
I don't know where it is/will be used, but making a comparison with the invoices list that is used in the redesign to show all the invoices, we only need some of the basics, I would remove the addresses, emails, telephones, website or just leave the default ones and show the others in the info. The emails have one that is primary, I guess we could return that one and if the department has more of them show it in the info

luisaterrabrito:
Please leave the currency, I think @lowiebenoot might be using it

lowiebenoot:
I'm indeed using the currencies on `departments.list`.

WouterSioen:
Who's following up to make the specs correct? @jenssegers ?

WouterSioen (dismissed):
I cleaned up the .list endpoint and got rid of a lot of fields.

Is this good to go now?

steftrenson (dismissed):

JordyDeGroeve (dismissed):

dietrichm (dismissed):

apiary.apib
```diff
@@ -291,6 +291,8 @@ Get a list of departments.
             + (object)
                 + id: `67c576e7-7e6f-465d-b6ab-a864f6e5e95b` (string)
                 + name: `Human Resources` (string)
+                + vat_number: `BE0899623035` (string)
+                + website: `https://piedpiper.com` (string)
```
dcsg: Q: why is this necessary in the `list`? My concern is that this field is not required and I would imagine that most of the times it will be `null`?

WouterSioen: I can imagine that you'd like to show that in an overview of departments?

WouterSioen: Okay, you're right, website is filled in for only 15% of departments and is not really prominent in the UI, so it'll be `null` a lot of the times.

Not sure if this is an issue though? I can still imagine this getting a more prominent place later?

WouterSioen: We could still add the fields at that moment though.

dcsg: The thing is if it's only 15% than the UI will have an empty column 85% of the times will be awkward for the user IMO other than that having the field would be ok ;)

dcsg (approved):

luisaterrabrito (approved):

paultimoce (approved):

WouterSioen:
Let's consider this proposal final so we can start on the implementation.

luisaterrabrito:
Sorry guys, I thought I was on my PR

PR #114, opened 2018-03-27
Proposal for events.invite

jenssegers:
Because adding invitees/attendees to an event can have side-effects, we should move it to its own action.


psrpinto:
👍 Nothing to declare. Let's not merge this though ;)

psrpinto:
We would also need an endpoint to remove attendees

jenssegers:
@regularjack Indeed, but could not find a decent name for that endpoint yet :) `uninvite`?

psrpinto:
`uninvite` would work. similar to `contacts.tag`/`contacts.untag`

WouterSioen (approved):

thomasnoe07:
@jenssegers only review needs to happen or something else as well?

thomasnoe07:
@jenssegers What can we do to ship this PR? Or can we close this?

psrpinto:
Closing this for now. It's the ideal way but we don't have plans to implement it at this point.

PR #132, opened 2018-04-23
[CORE-864] Companies.list primary_address

JordyDeGroeve:
Update **companies.list** to show both `primary_address` and `address`.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -773,7 +773,6 @@ Get a list of companies.
                         + email: `info@piedpiper.eu` (string)
                 + telephones (array[CompanyTelephone])
                 + website: `https://piedpiper.com` (string)
-                + address (Address) - Will be replaced by 'primary_address'.
                 + primary_address (Address)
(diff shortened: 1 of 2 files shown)


jenssegers (dismissed):

WouterSioen:
Not sure if we should keep the address documented since we will remove it and don't want to let people use it anymore?

JordyDeGroeve:
Wasn't sure if I should already remove it now or when we remove the implementation.
How do we handle these name chances in the docs, @jenssegers @stivni @mrnnn ?

WouterSioen (dismissed):
Let's not keep deprecated features documented, otherwise people will expect them to stay available.

jenssegers:
You're right @WouterSioen, let's remove it!

WouterSioen (approved):

mansilaxel (approved):

dietrichm (approved):

PR #133, opened 2018-04-23
events.cancel

psrpinto:


Makefile
```diff
@@ -1,6 +1,6 @@
 PWD = $(shell pwd)
 
-all: lint build
+all: lint build preview
```
jenssegers: Can we revert this? I never use preview, not really needed as a default make target

jenssegers: Can we revert this please?

jenssegers (dismissed):

Makefile
psrpinto: I'll revert but I think it makes sense to also generate the preview. For me at least, I test it locally so I always have to run make twice.

 :ok: by @regularjack on 24/04/2018 11:13:51

psrpinto:
@jenssegers Can you re-review? Addressed your remark

mrtus (approved):

Jdww (approved):

jenssegers (approved):

PR #137, opened 2018-04-25
Allow us to add custom fields to a new company

WouterSioen:


apiary.apib
```diff
@@ -870,6 +870,22 @@ Add a new company.
         + responsible_user_id: `2b12bd5d-d2d9-43ac-8755-36a1e00bd67f` (string, optional)
         + remarks: `Met at expo` (string, optional)
         + tags: prospect, expo (array[string], optional)
+        + custom_fields (array, optional)
+            + (object)
+                + id: `bf6765de-56eb-40ec-ad14-9096c5dc5fe1` (string)
+                + One Of
+                    + value: `092980616` (string)
+                    + value: `123` (number) - For integer, number and auto-increment fields
```
jenssegers: I think we're missing `Money` (amount + currency) too here.

WouterSioen: Looks like we don't allow people to pass a currency to money custom fields yet in the user interface and we just have a number here. 🤔 

 :ok: by @WouterSioen on 26/04/2018 09:53:19

JordyDeGroeve (dismissed):

JordyDeGroeve (approved):

apiary.apib
dietrichm: PP: We should just reuse the data structure (and add money there):
```
            + custom_fields (array[CustomField])
```

dietrichm (changes requested):

dietrichm (approved):
Make that a PP, that data structure is for field definitions. But a new one would be nice. :)

WouterSioen:
@dietrichm I didn't add a new data structure because it's the first place where it's used. Imo, data structures are only useful if they are re-used.

mansilaxel (approved):

apiary.apib
```diff
@@ -870,6 +870,22 @@ Add a new company.
         + responsible_user_id: `2b12bd5d-d2d9-43ac-8755-36a1e00bd67f` (string, optional)
         + remarks: `Met at expo` (string, optional)
         + tags: prospect, expo (array[string], optional)
+        + custom_fields (array, optional)
+            + (object)
+                + id: `bf6765de-56eb-40ec-ad14-9096c5dc5fe1` (string)
+                + One Of
+                    + value: `092980616` (string)
+                    + value: `123` (number) - For integer, number, money and auto-increment fields
+                    + value: `foo`,`bar` (array[string]) - For multiple selection fields
```
jenssegers: Missing `Money` here too?

jenssegers (approved):

PR #138, opened 2018-04-30
Add Money custom field

jenssegers:
When returning a monetary value in custom fields, it would be best to also return the currency of the account.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2519,6 +2519,7 @@ Their only use is to allow older integrations to migrate their data to the new A
     + value: `123` (number) - For integer, number and auto-increment fields
     + value: `foo`,`bar` (array[string]) - For multiple selection fields
     + value: true (boolean) - For Yes/No fields
+    + value (Money) - For money fields
     + value (object) - For related Teamleader objects
         + type: `company` (enum[string])
             + Members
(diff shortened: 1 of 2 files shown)


mrtus (approved):
Ah yes makes sense!

marijnvandevoorde (approved):

steftrenson (approved):

JordyDeGroeve (approved):

WouterSioen:
I did put this on hold since it's not implemented like this yet.

WouterSioen (approved):

WouterSioen:
@jenssegers I guess this should be changed for custom field values (when adding/updating entities with related custom fields too)?

Might be best to make a CustomFieldValue datastructure then?

dietrichm (approved):

WouterSioen:
Talked this through with Jens and we thin it might be better to not pass the currency at the moment since you can't change it.

jenssegers:
@WouterSioen Take the account currency for now.

jdrieghe (approved):

WouterSioen:
Merged in master to fix linting.

thomasnoe07:
@WouterSioen Should this PR stay open or can it be closed?

WouterSioen:
We can't really do that anymore, right, since this would be a BC breaking change? @jenssegers

stivni:
@jenssegers @dietrichm is this something we can pick up again?
I would actually like this, but if we don't find the time to implement this (BC) we should close this?

JordyDeGroeve:
Any update on this?

dietrichm:
Up to @jenssegers but I'd close this unless you deem this is in scope of some of the work you're going to pick up for the upcoming leap.

jenssegers:
Will close. Looks like it won't be picked up in the roadmap any time soon.

PR #140, opened 2018-05-02
Group customer and for attention of into invoicee on invoices.draft and update

steftrenson:


luisaterrabrito (dismissed):

jdrieghe (dismissed):

lowiebenoot (approved):

src/05-invoicing/invoices.apib
```diff
@@ -215,17 +220,18 @@ Update an invoice.
...
-                    + company
-            + id: `2cc40a9c-e674-43db-9b15-447ccb478200` (string)
+        + invoicee (object, required)
+            + customer (object, required)
+                + type (enum[string])
+                    + Members
+                        + contact
+                        + company
+                + id: `ebafa4c5-fa8a-4409-92e5-1b192243372f` (string)
+            + for_attention_of (object, required)
```
lowiebenoot: Shouldn't this be optional?

lowiebenoot (dismissed):

src/05-invoicing/invoices.apib
luisaterrabrito: In the update for now everything is required

WouterSioen: So if you create an invoice without a `for_attention_of`, you can't update it without adding one?

I guess this should be or optional or nullable?

jdrieghe: Right now I'm thinking `nullable` ... but then I think we should do the same in the create and make it required but nullable?

WouterSioen (dismissed):

lowiebenoot (approved):

luisaterrabrito (approved):

WouterSioen (approved):

PR #149, opened 2018-05-08
Add a related_status to the dealPhases.list endpoint

WouterSioen:
This is needed because otherwise integrators don't know which phase is
related to a certain status.

An example: team mobile does a `deals.win` and want to know to which
phase id this will resolve. If they want to do it right now, they have
to refetch the deal using deals.info afterwards.


DavidTeamleader (approved):

jenssegers (changes requested):
I don't think we should expose this to the outside. In the future we don't want phases to be related to a deal being won or lost. We want to be able to have won or lost deals, while remaining in the exact phase they were won or lost.

PR #150, opened 2018-05-08
Allow partial updates for invoices.update

jdrieghe:
### Changed
- `/invoices.update` now allows partial updates by making every root level item optional.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1927,7 +1927,7 @@ Update an invoice.
 
     + Attributes (object)
         + id: `b7023c11-455e-4fa5-bb96-87f37dbc7d07` (string, required)
-        + invoicee (object, required)
+        + invoicee (object, optional)
             + customer (object, required)
                 + type (enum[string])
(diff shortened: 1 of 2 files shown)


mrtus (approved):

WouterSioen (approved):

WouterSioen:
Some other Q's though about a consistency between add and update imo.

Shouldn't invoice_date be `nullable` in update, since you can add an invoice without an invoice_date. It should be possible to clear it, right?

Also: discounts doesn't have a required or optional in the API docs (for add), so not sure how that works?

steftrenson:
Making it nullable will just put the date on today. We might want to add this in the future, but for now I don't think it's necessary yet.

steftrenson (approved):

luisaterrabrito (approved):

jenssegers (approved):

jdrieghe:
This was deployed to production

PR #154, opened 2018-05-14
Add required scopes to API endpoints

jenssegers:
The API documentation is missing information about required scopes for our endpoints.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -113,13 +113,17 @@ Common used status codes:
  - `429` - Too Many Requests, your client has reached the API rate limit
  - `500` - Internal Server Error, something went wrong on our end
 
-## OAuth 2
+## Authentication
+
+### OAuth 2
 
 OAuth 2 is an authorization framework that allows a user to grant limited access to data in a Teamleader account, without having to expose their credentials.
 
 To get access to the protected resources using our API, OAuth 2 uses access tokens. An access token is a string representing the granted permissions. Access tokens can be obtained after a user has completed the OAuth 2 authorization flow.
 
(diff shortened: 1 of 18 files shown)


WouterSioen (dismissed):
What about related data? (F.e. customer coupled to a deal)

jdrieghe (dismissed):
@WouterSioen with quite an interesting question here

apiary.apib
```diff
@@ -2232,6 +2255,10 @@ Get a list of available tax rates.
                     + type: `department` (string)
                     + id: `182af0a8-3f68-409b-8941-cf8caf8f28a0` (string)
 
+## Payment Terms [/paymentTerms]
+
+*Required scopes: `invoices`*
```
stivni: I think this should be in CRM... because you can set a paymentTerm per customer without having invoicing

stivni:
@WouterSioen not an issue, as we don't return the data, but only a reference to it...
whenever we start side-loading, we will have to check if the user and/or the integration are allowed to

apiary.apib
jdrieghe: And what would you do with the paymentTerm besides create invoices?
I think the concept of a paymentTerm is only relevant for the invoicing module, so it's pointless to use it without the invoices scope.

steftrenson (dismissed):

apiary.apib
jenssegers: We can also drop this scope.

jenssegers:
I removed the scope for payment terms.

WouterSioen (approved):

steftrenson (approved):

JordyDeGroeve (approved):

PR #157, opened 2018-05-17
Add/modify resource descriptions

jenssegers:


apiary.apib
```diff
@@ -433,6 +433,8 @@ Get details for a single user.
 
 ## Custom Fields [/customFieldDefinitions]
 
+Custom fields are used to add additional data/properties to entities within Teamleader.
```
WouterSioen, line 436: Q: `entities` might be a bit of a technical term here? Wouldn't something like "other objects" be clearer?

It's API documentation though, so understanding what is meant with entities might not be an issue.

apiary.apib
```diff
@@ -1134,6 +1137,8 @@ Get the names of business types (legal structures) a company can have within a c
 
 ## Tags [/tags]
 
+Tags are used to label contacts or companies, so that they can be easily filtered or grouped into segments.
```
WouterSioen: PP: `so that they can easily be filtered` reads a bit better imo (swap easily and be)

apiary.apib
```diff
@@ -1166,6 +1171,8 @@ Get a list of tags.
 
 ## Deals [/deals]
 
+Deals are sales opportunities, which need to be followed up by their responsible user. A deal can move through specific deal phases, and be marked as won or lost.
```
WouterSioen: `sale opportunities` (singular)

apiary.apib
```diff
@@ -1437,7 +1444,9 @@ Get a list of lost reasons for deals.
                 + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string)
                 + name: `Deal declined` (string)
 
-## Deal Phases [/]
+## Deal Phases [/dealPhases]
+
+Deal phases are the different phases a deal typically goes through. A deal can be moved to either a next or previous phase. A deal can be marked as won or lost independently from the deal phase.
```
WouterSioen, line 1449: `A deal can be marked as won or lost independently from the deal phase.`

That's not true (yet)

apiary.apib
```diff
@@ -2305,8 +2316,7 @@ Get a list of product categories.
 
 ## Projects [/projects]
 
-Projects allow users to work together as a team on a single project.
-They comprise of one or more milestones, have one or more participants and are limited in time and budget.
+Projects allow users to work together as a team on a single project. They comprise of one or more milestones, have one or more participants and are limited in time and budget.
```
WouterSioen: `comprise` has a bit of a bad connotation (or that's at least how it's feels to me). I think `consists of` might be better here?

apiary.apib
```diff
@@ -2632,6 +2641,8 @@ Update a milestone.
 
 ## Time Tracking [/timeTracking]
 
+Time tracking gives you insight into the amount of work done and the work type. You are able to follow-up on the budgets of your projects and eventually you are able to invoice the billable time.
```
WouterSioen: `and the work type` might be better as `and its work type`

WouterSioen: > There is a space between follow and up under one condition—you are using follow up as a verb. To follow up means to gather further information or to reinforce or evaluate a previous action.

apiary.apib
jenssegers: :ok:

apiary.apib
jenssegers: :ok:

apiary.apib
jenssegers: :ok:

apiary.apib
jenssegers: I asked Niels to improve this text.

apiary.apib
```diff
@@ -2305,8 +2316,7 @@ Get a list of product categories.
 
 ## Projects [/projects]
 
-Projects allow users to work together as a team on a single project.
-They comprise of one or more milestones, have one or more participants and are limited in time and budget.
+Projects allow users to work together as a team on a single project. They consists of one or more milestones, have one or more participants and are limited in time and budget.
```
WouterSioen: `they consist of` (singular) 🙈

jenssegers: :ok:

apiary.apib
WouterSioen: @jenssegers I don't think this should've been merged with this information in it since this is incorrect...

jenssegers: @woutersioen I'll check with Niels for a new version!

PR #161, opened 2018-05-30
Add invoices.download documentation

jenssegers:


apiary.apib
```diff
@@ -1943,6 +1943,22 @@ Get details for a single invoice.
...
+        + id: `d885e5d5-bacb-4607-bde9-abc4a04a901b` (string, required)
+        + format: `pdf` (enum[string], required)
+            + Members
+                + pdf
+
++ Response 307 (application/json;charset=utf-8)
+
+    + Headers
+
+              Location: The url of the downloadable file. This will remain valid for 30 seconds.
```
WouterSioen, line 1960: Q: the indentation looks quite weird compared to the other endpoints. Is this intended?

WouterSioen (approved):

apiary.apib
jenssegers: https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md#example-18

WouterSioen: TIL

 :ok: by @WouterSioen on 30/05/2018 10:50:55

src/05-invoicing/invoices.apib
```diff
@@ -161,6 +161,22 @@ Get details for a single invoice.
...
 
+### invoices.download [POST /invoices.download]
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `d885e5d5-bacb-4607-bde9-abc4a04a901b` (string, required)
+        + format: `pdf` (enum[string], required)
+            + Members
+                + pdf
```
rathesDot, line 172: Do I understand that right, that we actually only offer `pdf` as format but leave it open to have more supported formats in the future?

rathesDot (approved):

src/05-invoicing/invoices.apib
WouterSioen: Yep, this is to ensure we can add formats without breaking backwards compatibility.

steftrenson (approved):

PR #181, opened 2018-06-25
Add documentation for timers.update endpoint

JordyDeGroeve:


WouterSioen (dismissed):

apiary.apib
```diff
@@ -2924,6 +2924,28 @@ Stop the current timer. This will add a new time tracking in the background.
...
             + id: `4ff475b2-f45b-4803-800e-350b9c02aeca` (string)
 
+### timers.update [POST /timers.update]
+
+Update the current timer. Only possible if there is a timer running.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + work_type_id: `db41328a-7a25-4e85-8fb9-830baacb7f40` (string, optional)
```
WouterSioen: Q: should this be nullable too? It's optional on the timer and time tracking objects, so I guess you should be able to clear this too.

rathesDot: So we want to be able to set all those values null in the timers endpoint?

WouterSioen: you can add a timer without those properties, so why wouldn't you be able to clear them if you did fill them in?

 :ok: by @rathesDot on 26/06/2018 12:42:38

WouterSioen (approved):

rathesDot (approved):

dietrichm (approved):

PR #193, opened 2018-07-19
Add hourly rate type to projects create

adognameddog:
### Important information about this pull request

Before we start implementing the actual functionality, we want some opinions and insights. The best way, we figured, was to show what we need through the API docs. I will not merge this PR, this is just to gather comments.

### What are we trying/going to do?

In order to further streamline the add project flow, the `projects.create` endpoint must be extended with the type of hourly rate. There are three distinct possibilities:

- hourly rate type "user": every user added to the project will use the default rate for that user from the settings.
- hourly rate type "task": every task in this project will use the default rate for that task from the settings.
- hourly rate type "fixed": each user, for each task, will use this hourly rate. In this case, the actual rate must also be supplied.

It will be possible to override rates for tasks and users, but only when editing the project.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2474,21 +2474,21 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
-        + billing_method (object, required)
+        + billing_method (object, optional)
             + One Of
                 + Properties
                     + method: `fixed` (string, required)
                     + amount (Money, required)
                 + Properties
-                    + method: `hourly_rate` (string, required)
-                    + hourly_rate (object, required)
+                    + method: `time_material` (string, required)
+                    + rate (object, required)
                         + One Of
                             + Properties
-                                + based_on: `user` (string, required)
+                                + `based_on`: `jezus_christus` (string, required)
                             + Properties
-                                + based_on: `work_type` (string, required)
+                                + `based_on`: `user` (string, required)
                             + Properties
-                                + based_on: `custom` (string, required)
+                                + `based_on`: `custom` (string, required)
                                 + rate (Money, required)
         + milestones (array, required) - At least one milestone is required
             + (object)
--- a/apiary.apib
+++ b/apiary.apib
@@ -2480,6 +2480,7 @@ Create a new project.
                     + method: `fixed` (string, required)
                     + amount (Money, required)
                 + Properties
+                    + method: `hourly_rate` (string, required)
                     + hourly_rate (object, required)
                         + One Of
                             + Properties
--- a/apiary.apib
+++ b/apiary.apib
@@ -2484,12 +2484,12 @@ Create a new project.
                     + rate (object, required)
                         + One Of
                             + Properties
-                                + `based_on`: `jezus_christus` (string, required)
-                            + Properties
                                 + `based_on`: `user` (string, required)
                             + Properties
+                                + `based_on`: `work_type` (string, required)
+                            + Properties
                                 + `based_on`: `custom` (string, required)
-                                + rate (Money, required)
+                                + amount (Money, required)
         + milestones (array, required) - At least one milestone is required
             + (object)
                 + due_on: `2018-01-01` (string, required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -2474,15 +2474,21 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
-        + hourly_rate (object, required)
+        + billing_method (object, required)
             + One Of
                 + Properties
-                    + type: `user` (string, required)
+                    + method: `fixed` (string, required)
+                    + amount (Money, required)
                 + Properties
-                    + type: `task` (string, required)
-                + Properties
-                    + type: `fixed` (string, required)
-                    + rate: `123` (number, required)
+                    + hourly_rate (object, required)
+                        + One Of
+                            + Properties
+                                + based_on: `user` (string, required)
+                            + Properties
+                                + based_on: `work_type` (string, required)
+                            + Properties
+                                + based_on: `custom` (string, required)
+                                + rate (Money, required)
         + milestones (array, required) - At least one milestone is required
             + (object)
                 + due_on: `2018-01-01` (string, required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -2474,6 +2474,15 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + hourly_rate (object, required)
+            + One Of
+                + Properties
+                    + type: `user` (string, required)
+                + Properties
+                    + type: `task` (string, required)
+                + Properties
+                    + type: `fixed` (string, required)
+                    + rate: `123` (number, required)
         + milestones (array, required) - At least one milestone is required
             + (object)
                 + due_on: `2018-01-01` (string, required)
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -2474,6 +2474,15 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + hourly_rate (object, required)
+            + One Of
+                + Properties
+                    + type: `user` (string, required)
+                + Properties
+                    + type: `task` (string, required)
```
stivni: B : `task` is something that doesn't exist anymore... shouldn't it be `work_type`?

apiary.apib
```diff
@@ -2474,6 +2474,15 @@ Create a new project.
...
+                + Properties
+                    + type: `fixed` (string, required)
+                    + rate: `123` (number, required)
```
stivni: B : a rate can never be just a number... it should probably be Money (amount & currency)

apiary.apib
```diff
@@ -2474,6 +2474,15 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + hourly_rate (object, required)
+            + One Of
+                + Properties
+                    + type: `user` (string, required)
```
stivni: PP : instead of `type`, should we go for something more descriptive?
`based_on`? or something similar?

apiary.apib
```diff
@@ -2474,6 +2474,15 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + hourly_rate (object, required)
```
stivni: Q : Is this what we charge the project's customer? Or is this our internal cost?
Shouldn't we make a clearer difference between those in our namings?

stivni: Also... this property doesn't really define the rate (only in 1 of the 3 cases...)... it defines something like the _pricing strategy_ or _invoicing method_ or ... so shouldn't we call it like that?
How will this thing be called in the UI?

src/07-projects/projects.apib
```diff
@@ -111,6 +111,22 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + billing_method (object, required)
```
psrpinto: B: I don't think we can make this required, it would be a BC break. Correct me if I'm wrong @stivni @jenssegers

src/07-projects/projects.apib
```diff
@@ -111,6 +111,22 @@ Create a new project.
...
+                    + method: `fixed` (string, required)
+                    + amount (Money, required)
+                + Properties
+                    + method: `time_material` (string, required)
+                    + hourly_rate (object, required)
+                        + One Of
+                            + Properties
+                                + based_on: `user` (string, required)
+                            + Properties
+                                + based_on: `work_type` (string, required)
```
psrpinto: B: Something's weird about the rendering here

screen shot 2018-07-23 at 14 24 10 1

psrpinto (changes requested):
👍  Blocking due to BC break

src/07-projects/projects.apib
WouterSioen: Nope, can't do

src/07-projects/projects.apib
```diff
@@ -111,6 +111,22 @@ Create a new project.
...
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + billing_method (object, required)
+            + One Of
+                + Properties
+                    + method: `fixed` (string, required)
+                    + amount (Money, required)
+                + Properties
+                    + method: `time_material` (string, required)
+                    + hourly_rate (object, required)
```
psrpinto: Q: Could we call this something that doesn't include the period? I'd rather not have `hourly_rate`, `weekly_rate`, etc, in the future, if we decide to implement that. Maybe just `rate`?

 :ok: by @regularjack on 24/07/2018 10:47:47

src/07-projects/projects.apib
adognameddog: It is required in the new add flow though, it will not be possible to not choose a billing method. I can of course make it optional in the API, but then we should figure out how we are going to handle that, because the front-end will need that.

src/07-projects/projects.apib
```diff
@@ -111,6 +111,22 @@ Create a new project.
...
+                        + One Of
+                            + Properties
+                                + based_on: `user` (string, required)
+                            + Properties
+                                + based_on: `work_type` (string, required)
+                            + Properties
+                                + based_on: `custom` (string, required)
+                                + rate (Money, required)
```
psrpinto: Q: Should we call this `amount` instead? Would be consistent with the naming when `method` is `fixed`

 :ok: by @regularjack on 24/07/2018 10:47:49

src/07-projects/projects.apib
WouterSioen: I guess the frontend should make fields required __before__ even sending data to the api?

 :ok: by @regularjack on 23/07/2018 17:50:44

src/07-projects/projects.apib
adognameddog: Because of an underscore in both "based_on" and "work_type", this happened. I escaped the property names as well.

 :ok: by @regularjack on 23/07/2018 17:50:48

psrpinto:
`apiary.apib` needs to be re-generated, it's not up-to-date.

adognameddog:
Thanks for the feedback. The actual work can start now. That doesn't mean nothing can or will change, we just needed the feedback.

psrpinto (approved):

apiary.apib
```diff
@@ -2474,6 +2474,22 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + billing_method (object, optional)
+            + One Of
+                + Properties
+                    + method: `fixed` (string, required)
+                    + amount (Money, required)
+                + Properties
+                    + method: `time_material` (string, required)
```
WouterSioen, line 2483: Does this mean `time or material`? And wouldn't it make more sense then to also name the property `time_or_material`? This looks a bit confusing to me now.

WouterSioen (approved):

apiary.apib
jdrieghe: It should actually be `time_and_materials` I think: https://en.wikipedia.org/wiki/Time_and_materials

jdrieghe: Free pro tip from the trenches: in absence of a real Domain Expert, Wikipedia is often your best friend.

apiary.apib
```diff
@@ -2474,6 +2474,22 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2016-02-04` (string, required)
+        + billing_method (object, optional)
```
jenssegers: Sorry for bringing up this discussion again, but to me `billing_method` sounds like it has something to do with visa, sepa or PayPal. Isn't there a better domain term for this?

apiary.apib
```diff
@@ -2474,6 +2474,22 @@ Create a new project.
...
+                + Properties
+                    + method: `time_material` (string, required)
+                    + rate (object, required)
+                        + One Of
+                            + Properties
+                                + `based_on`: `user` (string, required)
+                            + Properties
+                                + `based_on`: `work_type` (string, required)
+                            + Properties
+                                + `based_on`: `custom` (string, required)
```
jenssegers, line 2491: Isn't custom now the same a fixed?

adognameddog:
I'm closing this PR. This was just a vehicle to get feedback and insights. So long and thanks for all the shoes.

PR #195, opened 2018-07-23
Add proposal for sending urls to the web page of an item.

WouterSioen:


apiary.apib
```diff
@@ -647,6 +647,7 @@ Get a list of contacts.
                 + tags: prospect, expo (array[string])
                 + added_at: `2016-02-04T16:44:33+00:00` (string)
                 + updated_at: `2016-02-05T16:44:33+00:00` (string)
+                + canonical_url: `https://app.teamleader.eu/contact_detail.php?id=2a39e420-3ba3-4384-8024-fa702ef99c9f`
```
mrtus: Q: Should we start using meta data as a root for these? For me this is meta data and does not belong to the model itself.

WouterSioen: added_at and updated_at is also metadata imo

mrtus: The model can know about its own details, like creation & update times. (Domain)
It should/can not know its own location in the application. (Application)

TBD offcourse :) that is why this is a proposal right? 😉

src/02-crm/contacts.apib
```diff
@@ -66,6 +66,7 @@ Get a list of contacts.
                 + tags: prospect, expo (array[string])
                 + added_at: `2016-02-04T16:44:33+00:00` (string)
                 + updated_at: `2016-02-05T16:44:33+00:00` (string)
+                + canonical_url: `https://app.teamleader.eu/contact_detail.php?id=2a39e420-3ba3-4384-8024-fa702ef99c9f`
```
stivni: PP : not sure if "canonical" makes sense in this context

Isn't his just the "application_url", "visit_url", "web_url", "page_url" or something?

apiary.apib
WouterSioen: Yes, we kind of need this implemented, but aren't sure yet where to put it. Summoning @stivni and @jenssegers here.

src/02-crm/contacts.apib
WouterSioen: Canonical is wat Dietrich suggested based on what Drupal names this, but canonical probably references to the SEO meta tag "canonical_url" in Drupal.

Any preference?

stivni: B: I thought we agreed on adding the url in a meta-data-like property either in 

1. `data: { }, metaData : { url: ... }`
2. `data: { _url: ... }`
3. `data: { @url: ... }`

And do we wrap inside a `links` attribute?

```
links {
    "self": "https://app.teamleader.eu/foo/bar/baz"
}
```

stivni: @jenssegers thoughts?

WouterSioen: I know we're not doing REST, but in REST there's the concept of HATEOS where all links are under a `links` property which all have a rel (relationship) and a href (the actual link). This is f.e. also used for pagination.

mrtus: @stivni As mentioned here: https://github.com/teamleadercrm/api/pull/195#discussion_r204417716 I do like option 1 in combination with the `links` with maybe a more explicit url name like you mentioned in your first comment 👍

jenssegers: Not sure if it belongs in the (REST) `links` section, that's more used for navigating through relations or pagination.

I don't know if `canonical_url` is the correct name, but I kinda like it. We just need to make sure this is actually the correct term our url.

I would choose to either go for the currently proposed:

```
data: {
  id: 123,
  canonical_url: "https://app.teamleader.eu/foo/bar/baz"
}
```

Or by marking it somehow as meta data:

```
data: {
  id: 123,
  @canonical_url: "https://app.teamleader.eu/foo/bar/baz"
}
```

With my main preference going to the first one.

I'm just afraid that once we start adding meta data to objects by prepending it with a `_` or `@`, we will end up with endless discussions about what belongs in meta data and what not.

apiary.apib
WouterSioen: @mrtus we can't really differentiate between a `meta` and a `data` object anymore, since in `.list` endpoints, all data is returned as an array in the `data` key.

We could put an array `meta` in the `data` itself, but this doesn't give a lot of advantages imo.

 :ok: by @WouterSioen on 25/07/2018 15:13:03

src/02-crm/contacts.apib
WouterSioen: We went through a discussion and checked how major apis handled this. We came to the conclusion that `web_url` in the `data` object would be the most logical solution. (a bit more information in the commit message).

dietrichm (approved):

WouterSioen:
@stivni @jenssegers can we get this merged if we don't get a better proposal?

thomasnoe07:
@WouterSioen @jenssegers @stivni so to be merged or not that is the question here :-) If we do not merge it lets close this PR?

WouterSioen:
Let’s merge it instead. It’s implemented behind a feature flag which can easily be rolled out. Just want a final approval here before shipping

dietrichm:
Superseded by https://github.com/teamleadercrm/api/pull/225.

PR #202, opened 2018-08-03
[CORE-1049] Contacts.list sort on date_updated

JordyDeGroeve:


WouterSioen (dismissed):

apiary.apib
```diff
@@ -608,9 +608,11 @@ Get a list of contacts.
                 + field (enum[string], required)
                     + Members
                         + name - Sorts alphabetically on last_name and first_name
+                        + date_updated - Sorts on date_update
```
WouterSioen: Wait, shouldn’t this be “updated_at”?

WouterSioen (approved):

dietrichm (approved):

apiary.apib
```diff
@@ -608,9 +608,11 @@ Get a list of contacts.
                 + field (enum[string], required)
                     + Members
                         + name - Sorts alphabetically on last_name and first_name
+                        + updated_at - Sorts on date updated
```
WouterSioen, line 611: I would've updated the comment too

apiary.apib
```diff
@@ -608,9 +608,11 @@ Get a list of contacts.
...
                 + order (enum[string], optional)
                     + Members
                         + asc
+                        + desc
```
jenssegers, line 615: This means that you can now also sort `desc` on `name`?

JordyDeGroeve: Correct. This is already handled in the search. Was a small addition.

mrtus (approved):

rathesDot (approved):

PR #203, opened 2018-08-06
All address fields are required

jenssegers:
Our API expects all address field to be sent on API calls, but our API doesn't reflect this. Either we loosen our input validation, or apply this documentation change 👌

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3028,9 +3028,9 @@ Translates "meeting", "call" and "task" into their respective activity type UUID
 + number: `092980615` (string, required)
 
 ## Address (object)
-+ line_1: `Dok Noord 3A 101` (string, optional)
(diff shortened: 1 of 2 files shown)


stivni:
Not sure anymore about this one...
Should we allow sending less?

🤔

Jdww (approved):

WouterSioen (approved):

jenssegers:
@stivni Sending less information would improve user-friendly-ness of the API, but would make it less explicit that this is some sort of "value-object".

WouterSioen:
Let's first get this shipped? We can always loosen it up later.

dietrichm (approved):

JordyDeGroeve (approved):

PR #204, opened 2018-08-07
Contacts.list: add sorting on added_at

JordyDeGroeve:


mrtus (approved):

WouterSioen (approved):

dietrichm (approved):
Q/PP: I think the comments are useless when it's just "sorts on x". I would only add a comment for the ones sorting on multiple fields, since that is not always obvious from the sort name.

PR #219, opened 2018-09-07
[CAL-820] Add docs for todos.info

VerstraeteBert:
Adding docs for todos.info
Made a new tab for todos; after calender and before invoicing.


apiary.apib
```diff
@@ -1801,6 +1801,36 @@ Get a list of all activity types.
...
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `879770c5-8a7f-0067-8c25-b4993bdb5163` (string, required)
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
```
adognameddog, line 1822: Q: Is the array necessary? Only one Todo can be returned right?

apiary.apib
```diff
@@ -1801,6 +1801,36 @@ Get a list of all activity types.
...
+            + (object)
+                + id: `879770c5-8a7f-0067-8c25-b4993bdb5163` (string)
+                + description: `Design the landing page` (string)
+                + activity_type (object)
+                    + type: `activityType` (string)
```
adognameddog, line 1827: This should be an actual activity type?

apiary.apib
Jdww: Looking at other .info end points I'd say you're right. No need for data (array)

apiary.apib
Jdww: I don't think so, in the future it will be possible to add your own 'activityType' which you can name whatever you want and can be identified with the uuid. You'd have to get the name of the activityType from the activityType end point

src/05-todos/todos.apib
```diff
@@ -0,0 +1,27 @@
...
+        + data (array)
+            + (object)
+                + id: `879770c5-8a7f-0067-8c25-b4993bdb5163` (string)
+                + description: `Design the landing page` (string)
+                + activity_type (object)
+                    + type: `activityType` (string)
+                    + id: `28158af3-2ef3-4178-ac8b-2542b1481af0` (string)
+                + completed: true (boolean)
+                + completed_at: `2018-02-04T16:00:00+00:00` (string)
+                + ends_at: `2018-02-06T18:00:00+00:00` (string)
```
psrpinto, line 26: Q: Should this be `due_at`? That's what it was called in https://github.com/teamleadercrm/api-future/blob/master/src/03-activities/todos.apib#L92

src/05-todos/todos.apib
```diff
@@ -0,0 +1,27 @@
...
+        + id: `879770c5-8a7f-0067-8c25-b4993bdb5163` (string, required)
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `879770c5-8a7f-0067-8c25-b4993bdb5163` (string)
+                + description: `Design the landing page` (string)
+                + activity_type (object)
```
psrpinto, line 21: B?: I don't think todos should have an activity type

psrpinto (commented):
I think we need to clarify with Bram what fields should be made available in the todos endpoints

adognameddog:
I am closing this PR (and leaving the branch open). We do not know where todo's are going to, and once it is in the API it is hard, nigh impossible to backtrack from it.

PR #222, opened 2018-09-12
Add expected payment method to invoices draft

luisaterrabrito:


rathesDot (dismissed):

jdrieghe (dismissed):

dcsg (dismissed):

steftrenson (dismissed):

apiary.apib
```diff
@@ -2048,6 +2048,16 @@ Draft a new invoice.
                         + percentage
                 + description: `winter promotion` (string, required)
         + note: `Invoice comments` (string, optional)
+        + expected_payment_method (object, optional)
+            + method: `sepa_direct_debit` (enum[string], required)
+                + Members
+                    + `sepa_direct_debit`
+                    + `credit_card`
+            + initiator: `supplier` (enum[string], required)
```
jenssegers: The required initiator feels a bit confusing to me. Can you do `sepa_direct_debit` with the initiator set to `customer`? Isn't the expected payment method always from the supplier perspective, or am I missing a specific use case?

steftrenson: The initiator is the person who initiates the payment itself. For sepa direct debit you can have a supplier initiated payment, which means you already have a mandate from that customer to let him pay automatically. But you can also expect him to pay with sepa direct debit when you don't have the mandate yet. In that case,  the customer still needs to authorise the payment, by giving a one-off mandate for instance. Both use cases could be supported by a payment integrator, so we need to be able to provide this information to them. 

The same usecases are applicable for creditcard. If you expect someone to pay with creditcard, it does not mean that the payment will happen automatically. In Italy, you need to state which payment method you expect in order to generate an electronic invoice. Not all of these payment methods are supplier initiated (ie. automatic).

jenssegers: Ok, that makes it more clear. Might be confusing for users reading our documentation too though. Would it be possible to make `initiator` optional? So that you don't have to figure out what it means if you don't want to use it?

koen-serry (commented):
I find the required flag on the initiator a bit weird in the sense that it follows the method. Ie. SDD is always merchant initiated while Card usually customer initiated is. 
And since you can derive it from the method why have it there at all.

luisaterrabrito:
@jenssegers suggested accepting the reference as null, what do you think @steftrenson and @dcsg?

jenssegers (approved):

dcsg (approved):

dcsg:
Like we discuss (me, Stef and Jens) I will close this PR in favor of this one: https://github.com/teamleadercrm/api/pull/249

PR #225, opened 2018-09-19
Add documentation on web_url

tiagobastos:
Added _web_url_ on _companies_, _contacts_ and _deals_

Related to: [#195](https://github.com/teamleadercrm/api/pull/195)

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -936,6 +936,7 @@ Get a list of companies.
                     + id: `9b99d8f8-183b-4c7b-8354-9b98b47a192e` (string)
                 + added_at: `2016-02-04T16:44:33+00:00` (string)
                 + updated_at: `2016-02-05T16:44:33+00:00` (string)
+                + `web_url`: `https://app.teamleader.eu/company_detail.php?id=96a38bbf-24ed-4083-8a5c-20db92aa471e` (string)
                 + tags: prospect, expo (array[string])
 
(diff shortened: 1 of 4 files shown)


tiagobastos (commented):
Without the *`* around the word _web_url_ the output would appear misformatted. 
Anyone knows why?

mrtus:
@tiagobastos What does it do without the url? Can you try that?

mrtus:
@tiagobastos Could it be only a local issue? Can you give the endpoint `/invoices.download` a look where we also show a url? You can just remove the `` ` ``

rathesDot:
I can reproduce the issue. The problem seems to be the underscore in the url, not in the underscore in `web_url` ...

In general, the `_` is a reserved character in MSON, so using backticks is fine here. https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#6-reserved-characters--keywords

tiagobastos:
Actually, if only the key or the value has the underscore its fine. But both having it creates the misformatted output. So the backticks in both are the way to go, i think.

dietrichm (approved):

rathesDot (approved):
According to the definition of MSON, backticks should be added whenever we use reserved characters like _ ... SO this is fine here

paultimoce (approved):
looking good

PR #231, opened 2018-10-16
User preferences

jenssegers:
It would be handy for our frontend services and third party integrations, to know what kind of date formatting and number formatting preferences a user has. Currently, this is not adjustable through your user settings, but it could be in the future. For now this will be a read-only property that is generated by the backend, based on the user language and account country.

While adding the `preferences` root property to the user object, I realised that the time zone and language could also be considered a preference. Eg. everything that has an impact on how things are displayed in Teamleader.

**This would be a breaking change**. Unless we can find a better name for the property/structure, I think we should not be afraid to make breaking changes.

---

The date format was based on this Wikipedia article: https://en.wikipedia.org/wiki/Date_format_by_country

```
yy – two-digit year, e.g. 96
yyyy – four-digit year, e.g. 1996
m – one-digit month for months below 10, e.g. 4
mm – two-digit month, e.g. 04
mmm – three-letter abbreviation for month, e.g. Apr
mmmm – month spelled out in full, e.g. April
d – one-digit day of the month for days below 10, e.g. 2
dd – two-digit day of the month, e.g. 02
ddd – three-letter abbreviation for day of the week, e.g. Tue
dddd – day of the week spelled out in full, e.g. Tuesday
```

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -332,24 +332,32 @@ Get the current authenticated user.
             + last_name: `Smith` (string)
             + email: `john@teamleader.eu` (string)
             + telephones (array[Telephone])
-            + language: `nl-BE` (enum[string])
-                + Members
-                    + da-DK
-                    + de-DE
-                    + en-GB
-                    + es-ES
-                    + fi-FI
-                    + fr-FR
-                    + it-IT
-                    + nb-NO
-                    + nl-BE
-                    + nl-NL
-                    + pl-PL
-                    + pt-PT
-                    + sv-SE
-                    + tr-TR
             + function: `Sales` (string)
-            + time_zone: `Europe/Brussels` (string)
+            + preferences (object)
+                + language: `nl-BE` (enum[string])
+                    + Members
+                        + da-DK
+                        + de-DE
+                        + en-GB
+                        + es-ES
+                        + fi-FI
+                        + fr-FR
+                        + it-IT
+                        + nb-NO
+                        + nl-BE
+                        + nl-NL
+                        + pl-PL
+                        + pt-PT
+                        + sv-SE
+                        + tr-TR
+                + time_zone: `Europe/Brussels` (string)
+                + date_format (object)
+                    + short: `dd/mm/yyyy` (string)
+                    + medium: `dd mmm yyyy` (string)
+                    + long: `dd mmmm yyyy` (string)
+                + number_format (object)
+                    + thousand_separator: ' ' (string)
+                    + decimal_separator: '.' (string)
 
 ### users.list [GET /users.list]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -356,7 +356,7 @@ Get the current authenticated user.
                     + medium: `dd mmm yyyy` (string)
                     + long: `dd mmmm yyyy` (string)
                 + number_format (object)
-                    + thousand_separator: ' ' (string)
+                    + thousands_separator: ' ' (string)
                     + decimal_separator: '.' (string)
 
 ### users.list [GET /users.list]
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -44,7 +44,7 @@ Get the current authenticated user.
                     + medium: `dd mmm yyyy` (string)
                     + long: `dd mmmm yyyy` (string)
                 + number_format (object)
-                    + thousand_separator: ' ' (string)
+                    + thousands_separator: ' ' (string)
                     + decimal_separator: '.' (string)
 
 ### users.list [GET /users.list]
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -20,24 +20,32 @@ Get the current authenticated user.
             + last_name: `Smith` (string)
             + email: `john@teamleader.eu` (string)
             + telephones (array[Telephone])
-            + language: `nl-BE` (enum[string])
-                + Members
-                    + da-DK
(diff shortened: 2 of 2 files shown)


steftrenson (dismissed):

apiary.apib
```diff
@@ -332,24 +332,32 @@ Get the current authenticated user.
...
+                        + fi-FI
+                        + fr-FR
+                        + it-IT
+                        + nb-NO
+                        + nl-BE
+                        + nl-NL
+                        + pl-PL
+                        + pt-PT
+                        + sv-SE
+                        + tr-TR
```
nickwaelkens, line 352: Q: Should `tlh-KL` be in here too?

nickwaelkens (dismissed):

apiary.apib
lowiebenoot: I wouldn't add it. It's there just for development. People shouldn't know about it.

lowiebenoot (dismissed):

lowiebenoot:
Q: as this is a breaking change: shouldn't this use api versioning?

apiary.apib
```diff
@@ -332,24 +332,32 @@ Get the current authenticated user.
...
-                    + nl-NL
-                    + pl-PL
-                    + pt-PT
-                    + sv-SE
-                    + tr-TR
             + function: `Sales` (string)
-            + time_zone: `Europe/Brussels` (string)
+            + preferences (object)
+                + language: `nl-BE` (enum[string])
+                    + Members
```
lowiebenoot, line 338: Maybe not 100% in scope with this PR, but:

I'm not sure if the list is actually correct. Most of the languages are not country specific in Teamleader.
Spanish for example, is just `es` and not `es-ES`.
We actually only have `fr-FR` and `nl-NL` ( where `fr` and `nl` actually resemble `fr-BE` and `nl-BE`).

It's also missing 2 values that are available in the UI (`fr` and `ca`).

apiary.apib
```diff
@@ -332,24 +332,32 @@ Get the current authenticated user.
...
+                        + pl-PL
+                        + pt-PT
+                        + sv-SE
+                        + tr-TR
+                + time_zone: `Europe/Brussels` (string)
+                + date_format (object)
+                    + short: `dd/mm/yyyy` (string)
+                    + medium: `dd mmm yyyy` (string)
+                    + long: `dd mmmm yyyy` (string)
+                + number_format (object)
```
psrpinto, line 358: Q: Could we instead have a preference for region, instead of `date_format` and `number_format`? Not sure it's worth having such granularity, my guess would be that users probably just want to set it to "Germany".

steftrenson: I like the granular approach more because it's more explicit and flexible. I'm guessing there could be a lot of possible regions, with its own preferences. An integration would then also need the knowledge about what regions need what formatting, the granular approach is self-documenting.

jenssegers: @regularjack, thing is that many front services will need to still translate that region to a valid date and number format. Either we need to have all of these different rules in every service and share that with every third party service so that everything is consistent, or we provide it in a clean way through the API so that there's no debate on the format.

apiary.apib
jenssegers: We chose follow the ISO "locale" standard here. We shouldn't return a mix of countries and locales.

The missing values are out of scope though, those can be addressed in a different PR.

apiary.apib
jenssegers: You guys know that this is a public repository right? 😂

jenssegers:
@lowiebenoot Yes, we will have to introduce a new API version for this.

ArnaudWeyts (dismissed):

rathesDot (dismissed):

JordyDeGroeve (dismissed):

luisaterrabrito (dismissed):

apiary.apib
```diff
@@ -332,24 +332,32 @@ Get the current authenticated user.
...
+                    + thousand_separator: ' ' (string)
```
dietrichm: PP: go for `thousands_separator`, this is more commonly used than the singular form.

dietrichm (dismissed):

dietrichm (approved):
🎉

apiary.apib
jenssegers: I would even say that is B instead of PP!

JordyDeGroeve (approved):

ArnaudWeyts (approved):

luisaterrabrito (approved):

apiary.apib
lowiebenoot: Ah yeah, it should indeed change!

lowiebenoot (approved):

rathesDot (approved):

tiagobastos (approved):

steftrenson (approved):

steftrenson (approved):

adognameddog (approved):

fowbi (approved):

dcsg (approved):

geoffreydhuyvetters (approved):

jdrieghe (approved):
Anything planned still with this PR @jenssegers ?

stivni:
@jenssegers what do we do with this?

jenssegers:
@stivni Unless a team has time to pick it up I would either close it or keep it as inspiration.

PR #236, opened 2018-11-06
Add expected payment method to invoices info

luisaterrabrito:


apiary.apib
```diff
@@ -2042,6 +2042,12 @@ Get details for a single invoice.
             + note: `'Some extra remarks about the invoice'` (string)
             + created_at: `2016-02-04T16:44:33+00:00` (string)
             + updated_at: `2016-02-05T16:44:33+00:00` (string)
+            + expected_payment_method (object, nullable)
```
jenssegers, line 2045: `nullable` should be removed here (and also for reference)

luisaterrabrito: why?

jdrieghe (approved):

dcsg (approved):

apiary.apib
luisaterrabrito: https://github.com/teamleadercrm/api/blob/master/GUIDELINES.md#empty-properties

jenssegers: We kind of have a mix of having `nullable` on some properties and not on others for read endpoints. Eg. the `note` property can also return `null` right? So I was wondering if we should only document `nullable` on write endpoints, or shall we also add it to read endpoints?

luisaterrabrito: on invoicing at least I think we have nullable in the reads when applicable

dcsg: If we have properties that can be `nullable` and the documentation does not reflect that then we need to fix it in the documentation IMO.

dcsg:
Like we discuss (me, Stef and Jens) I will close this PR in favor of this one: https://github.com/teamleadercrm/api/pull/249

PR #237, opened 2018-11-06
projects.profit endpoint

adognameddog:
The data making up the profit for a project, is more sensitive than other project data, so we created a new endpoint to group that data. For now, we just return the total revenue of the project.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2592,6 +2592,21 @@ Get details for a single project.
                             + decision_maker
                             + member
 
+### projects.profit [GET /projects.profit]
+
+Get the data that makes up the profit of a single project: revenue, costs and profit.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `8a04371b-2ffb-407b-9b24-d5b5452009c7` (string, required)
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (object)
+            + id: `8a04371b-2ffb-407b-9b24-d5b5452009c7` (string)
+            + total_revenue (Money)
 
 ### projects.create [POST /projects.create]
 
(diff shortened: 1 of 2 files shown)


ketlane (approved):

jdrieghe (approved):
Not sure about the name "profit" for the endpoint, but looks good otherwise

WouterSioen (commented):
Any reason this is not returned in projects.info? As long as it's only the revenue, this endpoint doesn't seem to make a lot of sense to me?

adognameddog:
@WouterSioen This endpoint will return three simple monetary values: total revenue, total costs and the subsequent profit. This data is deemed sensitive for employees without full project access. That's why we can't use the `projects.info` endpoint, because then the endpoint would return different results depending on the user.

WouterSioen:
@adognameddog Should this be documented in the description of the endpoint? To make sure that people don't expect to have access to it by default?

jenssegers:
I'm having some mixed feelings about this endpoint. If we would start making new endpoints for every field that is conditionally returned based on a certain permission, that could make our API a bit messy.

This is the first use case where we have a field that would only be available for some users. Maybe it's not a bad thing to document that field as "only returned for users with permission X"?

Also most of our endpoints follow a more action based pattern such as `resources.action`. In this case `profit` is not really an action. Would `projects.report` be a better name?

I summon the great @stivni to join the conversation!

adognameddog:
@WouterSioen If we keep it like this, than the permission should certainly be mentioned in the API. I'll fix that.

@jenssegers That is certainly an option. We should make a decision soon though, since this is information is paramount for continuing the work on this. Thanks for the input!

adognameddog:
I suggest we keep this as-is for now. This should only be merged when the profit feature is released. by then, we might have different ideas for this endpoint.

So I'm not disagreeing with the feedback, but we need the revenue now, and then the total costs, and then the profit. So I'm going to use this endpoint for now.

So *do not* merge this. The firm thanks you.

PR #242, opened 2018-11-13
Add withholding taxes to invoices info

luisaterrabrito:


apiary.apib
```diff
@@ -2039,6 +2039,8 @@ Get details for a single invoice.
                                 + tax_exclusive_before_discount (Money)
                                 + tax_inclusive (Money)
                                 + tax_inclusive_before_discount (Money)
+                            + withholding_tax (object)
```
dcsg: Q/B: this should be nullable, right?

apiary.apib
```diff
@@ -2049,6 +2051,11 @@ Get details for a single invoice.
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+                + withholding_taxes (array)
```
dcsg: Will be weird for the countries where it does not apply to have this in the response.

dcsg (changes requested):

dcsg (dismissed):

luisaterrabrito:
Actually, implementing this, I noticed the vat rate is presented as zero not null. The withholding tax might not apply to an invoice in case the country is not Italy, but we cannot check that for an item. So should the withholding tax return the tax as zero either way?
@dcsg @jenssegers @steftrenson @jdrieghe

dcsg:
IMO if the user didn't set the withholding tax than it should not be zero and the object should not exist or the `withholding_tax` is null

tiagobastos (approved):

steftrenson (approved):

dcsg (approved):

PR #245, opened 2018-11-21
[CORE-540] adding the context filter to the custom field listing

paultimoce:
The context filter on custom field listing API endpoint was released some time ago but was not yet documented.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
 
     + Attributes (object)
         + page (Page, optional)
+        + filter (object, optional)
+            + context (enum[string], optional)
+                + Members
+                    + contact
+                    + company
+                    + form
+                    + event
+                    + product
+                    + project
+                    + milestone
+                    + sale
+                    + project_thirdparty
+                    + invoice
+                    + inv_invoice_item
+                    + subscription
+                    + meeting_report
+                    + meeting
+                    + todo
+                    + callback
+                    + ticket
+                    + pro_external_cost
+                    + werkbonnen
         + sort (array, optional)
             + (object)
                 + field (enum[string], required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -523,23 +523,13 @@ Get a list of all the definitions of custom fields.
                 + Members
                     + contact
                     + company
-                    + form
-                    + event
-                    + product
+                    + deal
                     + project
                     + milestone
-                    + sale
-                    + project_thirdparty
+                    + product
                     + invoice
-                    + inv_invoice_item
                     + subscription
-                    + meeting_report
-                    + meeting
-                    + todo
-                    + callback
                     + ticket
-                    + pro_external_cost
-                    + werkbonnen
         + sort (array, optional)
             + (object)
                 + field (enum[string], required)
--- a/src/01-general/custom-fields.apib
+++ b/src/01-general/custom-fields.apib
@@ -10,6 +10,28 @@ Get a list of all the definitions of custom fields.
 
     + Attributes (object)
         + page (Page, optional)
+        + filter (object, optional)
+            + context (enum[string], optional)
+                + Members
+                    + contact
+                    + company
+                    + form
+                    + event
+                    + product
+                    + project
+                    + milestone
+                    + sale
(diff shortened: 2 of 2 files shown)


apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+            + context (enum[string], optional)
+                + Members
+                    + contact
+                    + company
+                    + form
+                    + event
+                    + product
+                    + project
+                    + milestone
+                    + sale
```
mrtus: B: This one should be `deal`

Some other fields are not correct either.

mrtus (changes requested):
I think we should fix our implementation first before releasing this. Half of these context fields are not correct or up to date.

/cc @stivni

apiary.apib
paultimoce: @mrtus - This is actually correct like is imo. Please check the for column in the crm_custom_fields table to see why

paultimoce: You can also take a look at the CustomFieldDefinitionContext value object in core

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
 
     + Attributes (object)
         + page (Page, optional)
+        + filter (object, optional)
+            + context (enum[string], optional)
+                + Members
+                    + contact
+                    + company
+                    + form
```
stivni: B : this is not an existing type as far as the API is concerned
this is deprecated and we don't want it in the new API

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+                    + event
```
stivni: B : this is not an existing type as far as the API is concerned
this is deprecated and we don't want it in the new API

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+                    + product
```
stivni: PP : this is not an existing type as far as the API is concerned
we still need to add endpoints for products, so it doesn't make sense to have the custom fields for it

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+                    + project
+                    + milestone
+                    + sale
+                    + project_thirdparty
```
stivni: B : this is not an existing type as far as the API is concerned
we are deprecating third parties in this form

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+                    + invoice
+                    + inv_invoice_item
```
stivni: B : do we even want custom fields on invoice items?

apiary.apib
```diff
@@ -518,6 +518,28 @@ Get a list of all the definitions of custom fields.
...
+                    + subscription
+                    + meeting_report
```
stivni: B : is not part of the API yet

paultimoce:
As discussed with @stivni, I added filtering documentation for the contexts that are already documented in the listing API endpoint

Will follow-up with a PR which will fix the implementation

WouterSioen (approved):

JordyDeGroeve (approved):

mrtus (changes requested):
Good changes, but blocking until the implementation is fixed.

tiagobastos (approved):

dietrichm (approved):

jenssegers (approved):

ketlane (approved):

marijnvandevoorde:
@mrtus can you unblock? Tnx!

WouterSioen:
@mrnnn this should not be unblocked yet, since the implementation is not fixed yet. (We did a TA though about it and will fix this quite soon)

thomasnoe07:
@dietrichm I see that this PR is still open although it is mentioned it will be fixed soon. Can you have a look if we can push this PR or not? tx

stivni:
👉 poking @dietrichm & @WouterSioen to see if we can still do something with this...

dietrichm:
This is going to be picked up within a new story in our backlog. 💪

PR #247, opened 2018-11-26
materials.list

adognameddog:
Materials are added to a project and they affect the profitability. This information is only available for superusers and project decision makers.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2838,6 +2838,68 @@ Update a milestone.
 
 + Response 204 (application/json;charset=utf-8)
 
+## Materials [/materials]
+
+Materials are added to a project and they affect the profitability. This information is only available for superusers and project decision makers.
+
+*Required scopes: `projects`*
+
+### materials.list [GET /materials.list]
+
+Get a list of project materials.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, required)
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `cfb4146d-06be-41f1-bb39-aa3c929c71dc` (string)
+                + project (object)
+                    + type: `project` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + name: `Software license` (string)
+                + unit (object)
+                    + cost (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                    + price (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                + quantity: `3.0` (number)
+                + total (object)
+                    + cost (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                    + price (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                + billable: true (boolean)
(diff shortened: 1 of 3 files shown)


nickwaelkens (approved):

src/07-projects/materials.apib
```diff
@@ -0,0 +1,61 @@
...
+                                + excluding
+                    + price (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                + billable: true (boolean)
+                + milestone (object)
+                    + type: `milestone` (string)
+                    + id: `0709fc3e-e918-00d5-a113-ba00a33102ca`
```
JordyDeGroeve, line 53: PP: specify that this is a string
(same for supplier id)

src/07-projects/materials.apib
```diff
@@ -0,0 +1,61 @@
...
+                + departement (object)
+                    + type: `department` (string)
+                    + id: `a2d6bc44-2cf6-01c5-9a3c-67cabc1f029b`
+                + supplier (object)
+                    + type: `company` (string)
+                    + id: `232bdeaf-d5ca-0d36-bb3d-f7b540d99275`
+                + created_at: `2018-11-26T16:44:33+00:00` (string)
+                + updated_at: `2018-11-27T18:23:12+00:00` (string)
```
JordyDeGroeve, line 61: PP: add an empty line at the end of a new file

JordyDeGroeve (approved):
Checked manually 👌

apiary.apib
```diff
@@ -2838,6 +2838,68 @@ Update a milestone.
 
 + Response 204 (application/json;charset=utf-8)
 
+## Materials [/materials]
+
+Materials are added to a project and they affect the profitability. This information is only available for superusers and project decision makers.
```
Jdww, line 2843: Q: Do we can them superusers to the outside world? I thought we were trying to get rid of that word.

apiary.apib
```diff
@@ -2838,6 +2838,68 @@ Update a milestone.
...
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                    + price (object)
+                        + Include Money
+                        + tax: `excluding` (enum[string])
+                            + Members
+                                + excluding
+                + billable: true (boolean)
+                + milestone (object)
```
Jdww, line 2891: Should this be (object, nullable)? Same for department, supplier, (created_at?) and updated_at.

apiary.apib
stivni: indeed, we call them `account administrators`

```suggestion
Materials are added to a project and they affect the profitability. This information is only available for account administrators and project decision makers.
```

stivni: ^ the above suggestion will not work, as this is generated code

src/07-projects/materials.apib
psrpinto: ^ and department. String seems to be the default value, but I agree it's better to make it explicit.

psrpinto (approved):
LGTM, but I agree with other reviewers' remarks

ArnaudWeyts (approved):

apiary.apib
```diff
@@ -2838,6 +2838,68 @@ Update a milestone.
...
+
+### materials.list [GET /materials.list]
+
+Get a list of project materials.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, required)
```
jenssegers, line 2855: It's a bit weird to have a `.list` endpoint with a required filter in my opinion. This would make me think that this data structure should be part of the `projects.info` endpoint similar to how invoice items are also part of an invoice.

jenssegers (changes requested):
Blocking until we discuss the required filter question.

PR #248, opened 2018-11-27
Add electronic invoicing address to customer modules

WouterSioen:
Note that we put this under invoicing preferences. We'll move payment
terms and preferred_currency also to the invoicing preferences in a BC
breaking change, but let's add electronic_invoicing_address already in
it's final destination instead of putting it in the root and then moving
it later. This will help the Italian integrators to swiftly release the
Italian invoicing integrations.


apiary.apib
```diff
@@ -782,6 +782,8 @@ Get details for a single contact.
                         + id: `40157f9d-1322-4f2e-b245-188efdc52cc2` (string)
             + language: `en` (string)
             + payment_term (PaymentTerm, nullable)
+            + invoicing_preferences (object)
+                + electronic_invoicing_address (string, nullable)
```
dcsg, line 788: Q: should this be `electronic_invoice_address`?

WouterSioen: Aren't we using electronic_invoicing_address as domain term everywhere? I copy pasted the term from the codebase 😉

dcsg: Indeed, somehow I thought we were using the one I asked 🙈

dcsg (dismissed):

jdrieghe (dismissed):

src/02-crm/companies.apib
```diff
@@ -116,6 +116,8 @@ Get details for a single company.
             + language: `nl` (string)
             + preferred_currency: `EUR` (string, nullable)
             + payment_term (PaymentTerm, nullable)
+            + invoicing_preferences (object)
```
dcsg, line 121: Q: it might be nullable?

WouterSioen: I think we should always return an object here to make sure people always get the same datatype.

steftrenson (dismissed):

luisaterrabrito (dismissed):

rathesDot (dismissed):

WouterSioen:
I'd approve if I could

rathesDot (approved):

jenssegers:
@WouterSioen Any idea when you would plan the follow-up? If I'm not mistaken, invoicing will also have a small breaking change soon. Could bundle these two together 👌

jenssegers (approved):

WouterSioen:
@jenssegers I didn't see any documentation passing by for another breaking change. I was planning to do it whenever I saw another BC break passing by 😉

luisaterrabrito (approved):

PR #249, opened 2018-11-27
Add the expected payment method to Invoices .draft, .update and .info

dcsg:


jenssegers:
@dcsg Can we try `+ One Of` here? We have had cases where `+ One Of` would mess up the documentation rendering, but only if it is nested too deep I think. With `+ One Of` this could look something like this:

```
    + Attributes (object)
    + One Of
        + Properties
            + method: `sepa_direct_debit` (enum[string])
                + Members
                    + `sepa_direct_debit`
                    + `credit_card`
                    + `direct_debit`
            + reference: `AB1234` (string, required)
        + Properties
            + method: `sepa_direct_debit` (enum[string])
                + Members
                    + `cash`
                    + `cheque`
                    + `bankers_draft`
                    + `bank_transfer`
                    + `payment_card`
            + reference: `AB1234` (string, optional, nullable)
``` 

https://apiblueprint.org/documentation/mson/specification.html#52-one-of-type

luisaterrabrito (dismissed):

luisaterrabrito (dismissed):

lowiebenoot (dismissed):

rathesDot (dismissed):

jenssegers (dismissed):

apiary.apib
```diff
@@ -2058,7 +2058,8 @@ Get details for a single invoice.
                     + payment (object)
                         + Include Money
             + payment_reference: `+++084/2613/66074+++` (string)
-            + note: `'Some extra remarks about the invoice'` (string)
+            + note: `Some extra remarks about the invoice` (string)
+            + expected_payment_method (ExpectedPaymentMethod, optional)
```
steftrenson: Q: Do we return optional properties in a .info endpoint. I think this needs to be nullable instead of optional

apiary.apib
```diff
@@ -2131,6 +2132,7 @@ Draft a new invoice.
                         + percentage
                 + description: `winter promotion` (string, required)
         + note: `Invoice comments` (string, optional)
+        + expected_payment_method (ExpectedPaymentMethod, optional)
```
steftrenson, line 2481: PP: make it nullable in the .draft as well

apiary.apib
```diff
@@ -2182,6 +2184,7 @@ Update an invoice.
                         + product_category_id: `e2314517-3cab-4aa9-8471-450e73449041` (string, optional)
         + invoice_date: `2016-02-04` (string, optional)
         + note: `Some comments about the invoice` (string, optional, nullable)
+        + expected_payment_method (ExpectedPaymentMethod, optional)
```
steftrenson: We should make this partially updatable as well, so it should be nullable as well.

dcsg: `nullable` here would mean you are removing the `expected_payment_method` if it exists. But is, it is missing nullable but IMO we should keep optional for the partial update.

nielsvandenb:
@jenssegers @dcsg can this PR be reviewed and merged? The expected payment method is already returned in the API, but not yet documented.

apiary.apib
luisaterrabrito: then should we make all optional fields in the draft nullable?

luisaterrabrito: B: the code doesn't accept null

luisaterrabrito (changes requested):
invoices.draft doesn't expect a null as the expected payment method in the code

dcsg:
@nielsvandenb this is not yet available to all customers. Please speak with @PieterReel regarding when/how to release this

luisaterrabrito (dismissed):

dcsg:
On hold since this was not fully released.

jenssegers (dismissed):

stivni:
@steftrenson a small poke... is this functionality live? should we merge the docs as is?

stivni:
@steftrenson a small poke... is this functionality live? should we merge the docs as is?

luisaterrabrito:
It is live and works as is documented

luisaterrabrito (approved):

lowiebenoot (approved):

steftrenson (approved):

jenssegers (approved):

PR #251, opened 2018-11-28
Update extended description in invoices info, draft and update

luisaterrabrito:
[Issue](https://teamleader.atlassian.net/browse/IN-1239)
Invoices.draft/update doesn't allow an empty string for the extended description, although invoices.info does give us back an empty string for the extended description if we provided null to the draft or update.

So the proposal is:
- invoices info returns extended_description as null when it is not available
- invoices draft and update accept filled extended_description, null and it is optional


lowiebenoot (dismissed):

rathesDot (dismissed):

jenssegers (dismissed):

src/changes.apib
```diff
@@ -48,6 +48,11 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
+
+- The property `extended_description` returns null if the field is empty in invoices.info.
+In invoices.draft and invoices.update the `extended_description` is optional and nullable.
```
jenssegers: Missing a `-` here?

jenssegers:
`extended_description` was required in draft/update before? Should we consider this a breaking change or not?

dcsg (dismissed):

luisaterrabrito:
It was optional in the update validator and the draft didn't validate it at all, but was not documented as optional

steftrenson (dismissed):

PR #252, opened 2018-11-28
Add bookyear to invoice numbers

lowiebenoot:
- documented the breaking change where we added the bookyear name to the invoice number.
- corrected the formatting in the example invoice numbers

We currently are not adding the bookyear to the invoice number, so this is in fact a bug. But treating it as a breaking change, as people might have implemented things with wrong invoice numbers.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -298,6 +298,12 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
+
+- Added the bookyear name to the `invoice_number` on `invoices.list` and `invoices.info`.
+  Earlier it was only returning the number, e.g. `"27"`.
+  Now it adds the name of the bookyear as well, e.g. `"2018 / 27`".
+
 #### 2018-10-30
 
 - The top-level property `customer` was removed on all deal endpoints.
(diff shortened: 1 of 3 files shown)


jdrieghe (approved):

WouterSioen (approved):

luisaterrabrito (approved):

rathesDot (approved):

dcsg (approved):

steftrenson (approved):

stivni:
Can we actually find out how integrators expect this?
As one field or two fields?

alternative:

```
{
    "number": "27",
    "book_year": "2018"
}
```

lowiebenoot:
I am an integrator and I expect `2018 / 27`, because that's the actual invoice number. It's also the number that is displayed on my document. `27` alone, actually doesn't have meaning.

Making it an object is also possible, but doesn't give much added value in my opinion. It might even be weird, because people can still construct different invoice numbers then (`2018 / 27`, `2018-27`), which will result in weird things.

stivni:
Looking at this, the spaces still seem weird to me... how do other feel about this? @steftrenson @jdrieghe @lowiebenoot @luisabritotl @dcsg @EOSullivanBerlin

luisaterrabrito:
It doesn't really bother me, but I guess we can spend less time creating the strings without them

lowiebenoot:
It's how they are displayed on the documents as well, that's why I did it this way (it's reusing a function that already existed, I assume it's the same function).

PR #254, opened 2018-11-29
Add a description on how to use nullable/optional properties in the requests

dcsg:
It's not clear how we should handle `nullable` properties in the requests. This PR intends to make it more clear.

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -180,7 +180,7 @@ Example:
 
 #### When not to use `null` in properties
 
-1. Whenever you are updating a property that is optional/nullable and has a value assigned and you don't want to remove that value.
+1. Whenever you are updating a property that is optional/nullable and has a value assigned you don't want to remove.
 
 #### Nullable vs Optional
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -184,7 +184,7 @@ Example:
 
 #### Nullable vs Optional
 
-When you set a property to `null` your intention is to remove any assigned value to it.
+When you set a property to `null`, your intention is to remove any assigned value to it.
 When a property is optional you don't need to send the key in your request. The value will remain unchanged.
 
 **Note:** `Nullable` applies to the value of the property and `Optional` applies to the key of that property in the payload.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -175,7 +175,7 @@ Example:
 
 #### When to use `null` in properties
 
-1. When you want to unassign the value of that property.
+1. When you want to update a property and unassign it's value.
 2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
 
 #### When not to use `null` in properties
@@ -189,9 +189,13 @@ When a property is optional you don't need to send the key in your request. The 
 
 **Note:** `Nullable` applies to the value of the property and `Optional` applies to the key of that property in the payload.
 
-#### Note on empty strings
+#### Note on Empty strings
 
 Accepting empty strings does not have any meaning. If the goal is to unassign the value of a property then you should use `null` instead of an empty string.
+
+#### Note on Collections
+
+When the property you are updating is a collection, you **MUST** send all elements you want to keep or modify otherwise they will be removed.
 
 ### Date and time
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -175,7 +175,7 @@ Example:
 
 #### When to use `null` in properties
 
-1. When you want to update a property and unassign its value.
+1. In update endpoints, you can use `null` to unassign the value of a certain property.
 2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
 
 #### When not to use `null` in properties
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -176,7 +176,7 @@ Example:
 #### When to use `null` in properties
 
 1. In update endpoints, you can use `null` to unassign the value of a certain property.
-2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
+2. In add or create endpoints, you can choose to use `null` or to not pass optional properties.
 
 #### When not to use `null` in properties
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -151,7 +151,7 @@ Example of a `400 Bad Request` error:
 
 Adding endpoints and properties are typically always non-breaking and backwards compatible. Modifying, renaming and removing endpoints and properties is allowed during the beta period. Once our API is released, we will tag our API with a new version for each breaking change.
 
-### Empty properties
+### Response Empty properties
 
 Object properties without a value **MUST**:
 
@@ -170,6 +170,26 @@ Example:
   }
 }
 ```
+
+### Request Nullable/Optional properties
+
+#### When to use the `null` in properties
+
+1. When you want to unassign the value of that property.
+2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
+
+#### When not to use the `null` in properties
+
+1. Whenever you are updating a property that is optional/nullable and has a value assigned and you don't want to remove that value.
+
+#### Nullable vs Optional
+
+When you set a property to `null` your intention is to remove any assigned value to it.
+However when a property is optional you might don't want to send it if it's not changed.
+
+#### Note on empty strings
+
+Accepting empty strings does not have any meaning, if the goal is to unassign the value of a property than you should use `null` instead of an `empty string`.
 
 ### Date and time
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -175,7 +175,7 @@ Example:
 
 #### When to use `null` in properties
 
-1. When you want to update a property and unassign it's value.
+1. When you want to update a property and unassign its value.
 2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
 
 #### When not to use `null` in properties
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -185,7 +185,9 @@ Example:
 #### Nullable vs Optional
 
 When you set a property to `null` your intention is to remove any assigned value to it.
-However when a property is optional you might don't want to send it if it's not changed.
+When a property is optional, you don't need to send the key in your request, and the value will be unchanged.
+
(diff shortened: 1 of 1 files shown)


GUIDELINES.md
```diff
@@ -171,6 +171,26 @@ Example:
...
+2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
+
+#### When not to use the `null` in properties
+
+1. Whenever you are updating a property that is optional/nullable and has a value assigned and you don't want to remove that value.
+
+#### Nullable vs Optional
+
+When you set a property to `null` your intention is to remove any assigned value to it.
+However when a property is optional you might don't want to send it if it's not changed.
```
dcsg: Would it make sense to explain that `optional` means absent of the property key in the payload?

tiagobastos (dismissed):

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
 }
 ```
 
+### Request Nullable/Optional properties
+
+#### When to use the `null` in properties
```
WouterSioen:
```suggestion
#### When to use `null` in properties
```

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
...
+
+1. When you want to unassign the value of that property.
```
WouterSioen: This should probably state that this is during `update` calls?

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
...
+2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
+
+#### When not to use the `null` in properties
```
WouterSioen:
```suggestion
#### When not to use `null` in properties
```

GUIDELINES.md
dcsg: We can make it more explicit yes. But I assumed that to unassign a value the resource must be created first (so it would be an update).

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
...
+### Request Nullable/Optional properties
+
+#### When to use `null` in properties
+
+1. When you want to unassign the value of that property.
+2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
+
+#### When not to use `null` in properties
+
+1. Whenever you are updating a property that is optional/nullable and has a value assigned and you don't want to remove that value.
```
dietrichm: PP:
```suggestion
1. Whenever you are updating a property that is optional/nullable and has a value assigned you don't want to remove.
```

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
...
+
+#### Nullable vs Optional
+
+When you set a property to `null` your intention is to remove any assigned value to it.
+When a property is optional, you don't need to send the key in your request, and the value will be unchanged.
```
dietrichm: PP:
```suggestion
When a property is optional you don't need to send the key in your request. The value will remain unchanged.
```

GUIDELINES.md
```diff
@@ -171,6 +171,28 @@ Example:
...
+
+**Note:** `Nullable` applies to the value of the property and `Optional` applies to the key of that property in the payload.
+
+#### Note on empty strings
+
+Accepting empty strings does not have any meaning, if the goal is to unassign the value of a property than you should use `null` instead of an `empty string`.
```
dietrichm: PP:
```suggestion
Accepting empty strings does not have any meaning. If the goal is to unassign the value of a property then you should use `null` instead of an empty string.
```

GUIDELINES.md
WouterSioen: Yes, but I think documentation should be as explicit as possible and not assume that people will make the links themselves in their head.

GUIDELINES.md
dietrichm: IIRC this is different for custom fields. When updating an entity with the `custom_fields` key, all the missing optional custom fields (other than auto increment ones) will be cleared. So you are obliged to pass the current values if you want to keep them. 

Although the custom field values are not defined as properties of the entity, I still think we might have to note this in the documentation here.

WouterSioen: It's basically like that when talking about collections, not only custom fields (f.e. telephones, addresses, invoice items,...)

dcsg: But a collection is the value not the key of the property, right? But probably would make sense to explicit say that in the guidelines. When updating collections if you don't provide all items that already exist they will be removed. What do you think?

dietrichm: That kinda makes sense.

dcsg: I added a note on collections. Let me know if you agree with it! 😉

tiagobastos (dismissed):

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
 }
 ```
 
+### Request Nullable/Optional properties
+
+#### When to use `null` in properties
+
+1. When you want to update a property and unassign it's value.
```
dietrichm: B:
```suggestion
1. When you want to update a property and unassign its value.
```

dietrichm (dismissed):

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
 }
 ```
 
+### Request Nullable/Optional properties
+
+#### When to use `null` in properties
+
+1. When you want to update a property and unassign its value.
```
WouterSioen:
```suggestion
1. In update endpoints, you can use `null` to unassign the value of a certain property.
```

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
...
+2. If the property is optional and you are creating it you might want to use `null` to avoid the `if`s burden.
```
WouterSioen:
```suggestion
2. In add or create endpoints, you can choose to use `null` or to not pass optional properties.
```

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
...
+
+#### When not to use `null` in properties
+
+1. Whenever you are updating a property that is optional/nullable and has a value assigned you don't want to remove.
+
+#### Nullable vs Optional
+
+When you set a property to `null` your intention is to remove any assigned value to it.
```
WouterSioen:
```suggestion
When you set a property to `null`, your intention is to remove any assigned value to it.
```

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
...
+When a property is optional you don't need to send the key in your request. The value will remain unchanged.
+
+**Note:** `Nullable` applies to the value of the property and `Optional` applies to the key of that property in the payload.
+
+#### Note on Empty strings
+
+Accepting empty strings does not have any meaning. If the goal is to unassign the value of a property then you should use `null` instead of an empty string.
```
WouterSioen, line 194: Q: do we need to specify this? We normally validate against this, right? (`null` or a non-empty string)

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
...
+
+#### Note on Collections
+
+When the property you are updating is a collection, you **MUST** send all elements you want to keep or modify otherwise they will be removed.
```
WouterSioen:
```suggestion
When the property you are updating is a collection, you **MUST** send all elements you want to keep or modify, otherwise they will be removed.
```

GUIDELINES.md
dcsg: I'm not sure that we do always validate against this. But lately, there has been some discussion regarding using empty strings in the values.

WouterSioen: Well, we should validate against it to make sure users of our api just can't do this. 

Preventing > documenting

dietrichm: The goal though is to make validation less strict, i.e. to accept empty strings in addition to `null`. This was discussed with @stivni in the context of the IBAN validation in the API. I will check to write a blog post on this.

WouterSioen: But do we handle empty strings as a `null` value then or as a value containing an empty string?

WouterSioen: (I vote for `null`)

steftrenson (approved):

luisaterrabrito (approved):

GUIDELINES.md
stivni: In most cases an empty string and `null` can be perceived as the same value, so if people send and empty string, but we don't want to store it like that, we should convert it first to `null` instead of not accepting it

WouterSioen: Can this decision be documented somewhere? (blogpost + meta repository f.e.?)

GUIDELINES.md
```diff
@@ -171,6 +171,32 @@ Example:
 }
 ```
 
+### Request Nullable/Optional properties
```
stivni, line 174: PP : I think you're making this very difficult to understand.

The general principle is the following.

If you don't provide a certain property in your request, the previous value will remain.
If you provide a certain property in your request, the provided value will overwrite the previous value for that property.

Hence sending `null` as a value, will empty the property.
Hence for sending a collection or an object, you need provide all items/properties.

Also: shouldn't this be in the html documentation instead of the contribution guidelines?

GUIDELINES.md
stivni: I'm not sure if we should explain an existing standard?

GUIDELINES.md
dcsg: > Also: shouldn't this be in the HTML documentation instead of the contribution guidelines?

Indeed, this makes more sense in the HTML documentation.

kevinwaelkens (approved):

GUIDELINES.md
jenssegers: The main purpose for this documentation section was to make it clear when to use `nullable` and `optional` which are very related to the design of the API and not the usage. So a section explaining this in guidelines makes a lot of sense.

How we handle empty values is something different and has its place in the general documentation indeed.

tiagobastos (approved):

adognameddog (approved):

r1dotl (approved):

lowiebenoot (approved):

stivni:
Sorry, I still don't understand what the goal of this PR is.
It seems like this is written towards api clients/user. But these guidelines are for API design.

If we want to document when to make properties `nullable` and when to make them `optional`, we should aim towards ourselves.

Feel free to make a second attempt though

PR #256, opened 2018-12-11
Add `outstanding` and `matched` statuses to `invoices.list` and `invoices.info`

dcsg:
The `booked` status has more than one meaning making it difficult for the integrators to distinguish when an `invoice` was `outstanding` or `matched`. We now do also allow to filter the invoices with multiple `statuses`.


apiary.apib
```diff
@@ -298,6 +298,10 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
```
steftrenson, line 330: Keep in mind that this will give conflicts when @lowiebenoot 's invoice number change will be merged.

apiary.apib
```diff
@@ -1910,10 +1914,11 @@ Get a list of invoices.
     + Attributes (object)
         + filter (object, optional)
             + department_id: `080aac72-ff1a-4627-bfe3-146b6eee979c` (string, optional) - The ID of the department you wish to filter on.
-            + status: `booked` (enum[string], optional) - Only return booked or draft invoices. Leave blank to return everything
+            + status: `matched` (enum[string], optional) - Only return booked or draft invoices. Leave blank to return everything
```
steftrenson: Can you also update the explanatory text?

luisaterrabrito (dismissed):

apiary.apib
lowiebenoot: When it will be merged...

apiary.apib
```diff
@@ -298,6 +298,11 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
+
+- The property `status` in `invoices.list` and `invoices.info` now returns `due` and `matched` instead of `booked`.
+- The `filter` `status` property in the `invoices.list` request now only accepts an `array` of values.
```
lowiebenoot: Can't we just accept both string and array of strings?

apiary.apib
```diff
@@ -1910,10 +1915,11 @@ Get a list of invoices.
     + Attributes (object)
         + filter (object, optional)
             + department_id: `080aac72-ff1a-4627-bfe3-146b6eee979c` (string, optional) - The ID of the department you wish to filter on.
-            + status: `booked` (enum[string], optional) - Only return booked or draft invoices. Leave blank to return everything
-                + Members
+            + status: `due`, `matched` (array[string]) - Empty array returns everything.
```
lowiebenoot: Leaving it blank (as in not passing the filter) should still return everything, right?

dcsg: leaving it blank does not make sense, simply don't send it in the filter and you will have everything 😉 which makes me things that the `Empty array returns everything` does not make sense, and this should also be optional 🤔

apiary.apib
dcsg: we can, but why you would do that? having two types to deal with brings complexity?

lowiebenoot: It will not break backward compatibility if you accept both.

dcsg: True, but we are breaking it on the split of the `booked` to `due` and `matched`. So breaking it here makes it more suitable and less complex IMO. The developers don't have to think about "if I want to filter only by one status should I send it in the array or just a string?".

luisaterrabrito (dismissed):

lowiebenoot (dismissed):

jenssegers:
What does due and matched mean exactly? Both are a bit unclear to me.

I think there might be a difference in the "status" that says if an invoice is draft/booked/declared and the payment-status or whatever you want to call it that says if a payments has been received, if it is overdue, etc.

From an integration perspective I think it makes sense to fetch all booked invoices, regardless if the invoice was paid or not. But from a UI perspective it would probably be the other way around.

I remember having this discussion in the beginning of the design of the invoicing API about state vs. status.

```
status: a situation at a particular time, especially in an argument, discussion etc.
state: the physical or mental condition that someone or something is in
```

Maybe the state of an invoice reflects if it is draft/booked/declared. And the status reflects if it is pending/late/paid/uncollectable/reminded.

Or would it make more sense to make this a `payment_status`?

dcsg:
@jenssegers after some discussion we came to the conclusion that `outstanding` better represents the status of the invoice instead of `due`. `due` and `overdue` are more of a payment status while `outstanding` is the status of an invoice as soon as you book it until it gets matched/fully paid.

apiary.apib
stivni: maybe we should make these PRs toward a "staged-breaking-changes" branch or something? and then we can merge that in master whenever we released the version?

apiary.apib
stivni: > The developers don't have to think about "if I want to filter only by one status should I send it in the array or just a string?".

But they don't have to think about that... they can just pass an array with 1 item.

The other way around: if you want to filter on more values, you will know it needs to be an array instead of a string...

So I would be fine with keeping both options for the sake of "simplicity for the simple use case", but no strong opinions about it.

apiary.apib
stivni: That makes sense from the point of the integrator... so they don't have to write conditionals.
Although this depends on the fact if we still want to support strings too.

dcsg: IMO supporting two different ways of doing it only brings complexity to both sides. I would rather keep it simple and if there is a real need I would support both.

nielsvandenb:
@dcsg can we move this PR forward? I need it :)

dcsg:
@nielsvandenb we're still working on the implementation of this.

src/changes.apib
```diff
@@ -48,6 +48,11 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
+
+- The property `status` in `invoices.list` and `invoices.info` now returns `outstanding` and `matched` instead of `booked`.
+- The `filter` `status` property in the `invoices.list` request now only accepts an `array` of values.
```
stivni: B : the filter also doesn't allow to send it "booked" as a status anymore... should we tell customers what to do instead? ("send `["oustanding", "matched"]` to keep filtering on "booked"?)

stivni (dismissed):
see comment

PR #257, opened 2018-12-12
[PRO-1251] Documentation for the project items list endpoint

ketlane:
Project items are added to a project and they affect the profitability. This information is only available for account administrators and project decision makers.


ArnaudWeyts (dismissed):

apiary.apib
```diff
@@ -2881,6 +2881,66 @@ Update a milestone.
...
+
+### projectItems.list [GET /projectItems.list]
+
+Get a list of project items.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, required)
```
adognameddog: I think this should be optional. That would mean it would be possible to retrieve all the project items for every projects you have access to (which is not a realistic use case), but I don't think that's a problem.

apiary.apib
```diff
@@ -2881,6 +2881,66 @@ Update a milestone.
...
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `cfb4146d-06be-41f1-bb39-aa3c929c71dc` (string)
+                + project (object)
+                    + type: `project` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + type: `time_tracked`
+                + description: `Software license` (string)
```
adognameddog: If you use "time_tracked" as the example, I would not use "Software license" as the description. Rather some realistic description for time tracking, like "Drafted the document" or "Research".

apiary.apib
```diff
@@ -2881,6 +2881,66 @@ Update a milestone.
...
+                    + type: `employee` (string)
+                    + id: `a2d6bc44-2cf6-01c5-9a3c-67cabc1f029b`
+                + milestone (object)
+                    + type: `milestone` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + unit_internal_cost (object)
+                    + Include Money
+                + unit_billable_price (object)
+                    + Include Money
+                + quantity: `3.0` (number)
```
adognameddog: PP: Use `3.25` so it is extra clear that quantity is a decimal number. It's pretty clear now, but using an actual value after the decimal point will make it even clearer (and it is realistic: 3 hours and 15 minutes)

JordyDeGroeve (dismissed):
Approve, but it would be nice to see Peter's feedback handled.

apiary.apib
ketlane: Ok, but let's first see what the outcome of the meeting around this parameter is.

apiary.apib
ketlane: Fixed

apiary.apib
ketlane: Fixed

adognameddog (dismissed):

ArnaudWeyts (dismissed):

apiary.apib
```diff
@@ -2881,6 +2881,70 @@ Update a milestone.
...
+                    + type: `milestone` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + type: `time_tracked` (enum[string])
+                    + Members
+                        + time_tracked
+                        + external_cost_item
+                + description: `Drafted the document` (string, nullable)
+                + date: `2018-11-26T16:44:33+00:00` (string)
+                + employee (object, nullable)
+                    + type: `employee` (string)
```
mrtus: B: The type should be `user` so lets fix that?

mrtus (changes requested):

apiary.apib
adognameddog: Right! Will fix.

ketlane: Fixed

apiary.apib
ketlane: Fixed

ArnaudWeyts (dismissed):

apiary.apib
```diff
@@ -2881,6 +2881,70 @@ Update a milestone.
...
+                + id: `cfb4146d-06be-41f1-bb39-aa3c929c71dc` (string)
+                + project (object)
+                    + type: `project` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + milestone (object, nullable)
+                    + type: `milestone` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + type: `time_tracked` (enum[string])
+                    + Members
+                        + time_tracked
```
jenssegers: Could also be `tracked_time` actually. Sounds a bit more natural.

apiary.apib
```diff
@@ -2881,6 +2881,70 @@ Update a milestone.
...
+                        + external_cost_item
```
jenssegers: Since it is already a project item, this can be `external_cost`?

src/07-projects/project-items.apib
```diff
@@ -0,0 +1,63 @@
...
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + milestone (object, nullable)
+                    + type: `milestone` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + type: `time_tracked` (enum[string])
+                    + Members
+                        + time_tracked
+                        + external_cost_item
+                + description: `Drafted the document` (string, nullable)
+                + date: `2018-11-26T16:44:33+00:00` (string)
```
jenssegers: What does this date represent?

apiary.apib
```diff
@@ -2881,6 +2881,70 @@ Update a milestone.
...
+                + description: `Drafted the document` (string, nullable)
+                + date: `2018-11-26T16:44:33+00:00` (string)
```
jenssegers: What does this date represent?

apiary.apib
adognameddog: Ah, no. Because the external cost has multiple items. Those items can be even be linked to no, or other, projects.

apiary.apib
adognameddog: The date most relevant for the project item. In case of time tracked: the date on which it was performed. For an order item: the date it was added.

adognameddog (dismissed):

src/07-projects/project-items.apib
```diff
@@ -0,0 +1,63 @@
...
+                + employee (object, nullable)
```
mrtus: As we try to avoid the term `employee` I think we should try to find another name. /cc @stivni @jenssegers

adognameddog: I think employee is the correct term here. It references the employee who did the time that was tracked, or the employee that added the order item.

mrtus: But we only want to use the term `employee` in HR context. That is why I reckon we should try to find another name.

stivni: we didn't get this straight enough yet, indeed...
can I propose to wait until the domain discovery workshop before we coin the terms?

apiary.apib
jenssegers: Can we find a better name than just `date` for his concept? We should aim to make our API design as easy to understand as possible without having to go to the documentation.

ketlane: Any proposals? :D I have 'item_date' / 'happened_at' / 'date_accomplished' / 'accomplished_at' / 'relevant_date' but I still feel just 'date' is better :)

src/07-projects/project-items.apib
mrtus: Looking at the current projects api documentation, we have `participants` there. So why aren't we calling this `participant` here as well? Atleast that would make to most sense to reuse that term in the project domain I reckon?

apiary.apib
ketlane: Ok, replaced.

src/07-projects/project-items.apib
ketlane: Ok, updated to participant.

apiary.apib
ketlane: Unless there's any other proposal, I will rename it to 'item_date'. This will also apply to sorting.

@jenssegers @adognameddog @mrtus @rathesDot @tiagobastos

mrtus: I don't think `item_date` explains so much what it is about, the only thing I can come up with is `date_of_occurence` but I don't think that is any better

mrtus (dismissed):

apiary.apib
ketlane: We decided to call it `occurred_at`

adognameddog (dismissed):

tiagobastos (dismissed):

ArnaudWeyts:
Won't `unit_internal_cost` and `internal_cost` also be nullable?

ketlane:
> Won't `unit_internal_cost` and `internal_cost` also be nullable?

These fields will be reflected as 0 if they are not set, so not nullable.

adognameddog (dismissed):

apiary.apib
```diff
@@ -2910,6 +2910,77 @@ Update a milestone.
...
+                        + tracked_time
+                        + external_cost_item
+                + description: `Drafted the document` (string, nullable)
+                + occurred_at: `2018-11-26T16:44:33+00:00` (string)
+                + participant (object, nullable)
+                    + type: `user` (string)
+                    + id: `a2d6bc44-2cf6-01c5-9a3c-67cabc1f029b` (string)
+                + unit_internal_cost (object)
+                    + Include Money
+                + unit_billable_price (object)
```
mrtus: Q: Currently, this naming is a bit weird and I reckon it would be much more readable to switch around some words. Should we call this `internal_unit_cost` instead? I think this would make more sense because we describe a cost per unit as `unit_cost`

(Same for `billable_unit_price`)

apiary.apib
```diff
@@ -2910,6 +2910,77 @@ Update a milestone.
...
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + milestone (object, nullable)
+                    + type: `milestone` (string)
+                    + id: `944534fb-15f1-4eea-aab1-82a427aa2d0d` (string)
+                + type: `tracked_time` (enum[string])
+                    + Members
+                        + tracked_time
+                        + external_cost_item
+                + description: `Drafted the document` (string, nullable)
+                + occurred_at: `2018-11-26T16:44:33+00:00` (string)
```
luisaterrabrito: what about created_at?

luisaterrabrito: Just saw the added_at, what is the difference?

ketlane: The occured_at field refers to the date the action happened (i.e. for the case of time tracking, it's the date chosen by the user as the beginning of the tracking period, while added_at is the time when the entry was added to the database)

apiary.apib
```diff
@@ -2910,6 +2910,77 @@ Update a milestone.
 
 + Response 204 (application/json;charset=utf-8)
 
+## Project items [/projectItems]
+
+Project items are added to a project and they affect the profitability. This information is only available for account administrators and project decision makers.
```
jenssegers: Can we check with product marketing for some copy here?

apiary.apib
```diff
@@ -2910,6 +2910,77 @@ Update a milestone.
...
+                + occurred_at: `2018-11-26T16:44:33+00:00` (string)
+                + participant (object, nullable)
+                    + type: `user` (string)
+                    + id: `a2d6bc44-2cf6-01c5-9a3c-67cabc1f029b` (string)
+                + unit_internal_cost (object)
+                    + Include Money
+                + unit_billable_price (object)
+                    + Include Money
+                + quantity: `3.25` (number)
+                + unit: `hour` (string, nullable)
```
jenssegers: Should this be an enum?

adognameddog: No, since every account can have their own units.

adognameddog (dismissed):

luisaterrabrito (dismissed):

apiary.apib
ketlane: Checked, they (Niels, Benny, Bram) want to keep this the same as it is now.

apiary.apib
ketlane: Discussed this with @stivni and @adognameddog and decided to use "internal_cost_per_unit" and "billable_price_per_unit"

adognameddog (dismissed):

luisaterrabrito (dismissed):

ArnaudWeyts (dismissed):

luisaterrabrito (dismissed):

tiagobastos (dismissed):

tiagobastos (dismissed):

tiagobastos (dismissed):

luisaterrabrito (dismissed):

src/apiary.apib
```diff
@@ -67,6 +67,8 @@ HOST: https://api.teamleader.eu
 
 :[Milestones](./07-projects/milestones.apib)
 
+:[Project Items](./07-projects/project-items.apib)
```
joaoinacio: As discussed we could  for now **not** include project-items in apiary.apib and merge this documentation.

joaoinacio (approved):

tiagobastos (approved):

adognameddog (approved):

adognameddog:
Manually checked 👍

ArnaudWeyts (approved):

nickwaelkens (approved):

PR #258, opened 2018-12-20
Collection pagination

ketlane:
This is a proposal for better documenting our pagination.

### What has been done
- Created a response pagination data structure that can be used in future PRs (see https://github.com/teamleadercrm/api/pull/257 for an example)
- Updated the intro to give an example of a response with pagination

### DIFF
--- a/src/datastructures.apib
+++ b/src/datastructures.apib
@@ -59,6 +59,12 @@
 + number: `1` (number, optional)
     + Default: `1`
 
+## Pagination
++ count: `200` (number)
++ per_page: `10` (number)
++ current_page: `2` (number)
++ total_pages: `20` (number)
+
 ## PaymentTerm  (object)
 + type (enum[string])
     + Members
--- a/apiary.apib
+++ b/apiary.apib
@@ -84,13 +84,11 @@ Sample response with pagination:
             "email": "john@piedpiper.eu"
         }
     ],
-    "_meta": {
-        "_pagination": {
-            "count": 200,
-            "per_page": 2,
-            "current_page": 1,
-            "total_pages": 100,
-        }
+    "pagination": {
+        "count": 200,
+        "per_page": 2,
+        "current_page": 1,
+        "total_pages": 100,
     }
 }
 ```
--- a/apiary.apib
+++ b/apiary.apib
@@ -65,6 +65,36 @@ Both single objects and collections are returned as JSON objects, keyed by `data
 
 Responses may also contain meta data for pagination.
 
+Sample response with pagination:
+```json
+{
+    "data": [
+        {
+            "id": "f1dfb84c-3c29-4548-9b9b-9090a080742a",
+            "first_name": "Erlich",
+            "last_name": "Bachman",
+            "salutation": "Mr",
+            "email": "info@piedpiper.eu"
+        },
+        {
+            "id": "f1dfb84c-3c29-4548-9b9b-9090a080742b",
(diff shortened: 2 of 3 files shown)


luisaterrabrito (dismissed):

jenssegers:
PP adjust to fit our current API style:

```
{
    data: [],
    included: [],
    pagination: {
        count: 1000,
        per_page: 10,
        current_page: 1,
        total_pages: 100 
    }
}
```

rathesDot:
Since when did we decide on how pagination is going to look like? Did I miss anything here?

luisaterrabrito:
The description says it is a proposal @rathesDot

rathesDot:
Oh, for me @jenssegers comment sounded like there is already a way to do pagination. Just misinterpreted it 😅 

EDIT: Agree on that counter-proposal by @jenssegers

tiagobastos:
Cool one! 👍 

Throwing ideas now. 
We could even apply HATEOAS here. 
I know it's for REST but the concept could be extended to this case. 
Something like the first half of this example [symfonycasts](https://symfonycasts.com/screencast/rest-ep2/pagination)

ketlane:
> Cool one! 👍
> 
> Throwing ideas now.
> We could even apply HATEOAS here.
> I know it's for REST but the concept could be extended to this case.
> Something like the first half of this example [symfonycasts](https://symfonycasts.com/screencast/rest-ep2/pagination)

Since most (all?) of our endpoints are POST requests, I don't think HATEOAS would help very much, though, no?

jenssegers (changes requested):
Blocking because not following current API style.

ketlane:
I have adjusted the example to fit the current API style.

luisaterrabrito (approved):

dcsg (approved):

rathesDot (approved):

src/datastructures.apib
```diff
@@ -59,6 +59,12 @@
 + number: `1` (number, optional)
     + Default: `1`
 
+## Pagination
++ count: `200` (number)
++ per_page: `10` (number)
++ current_page: `2` (number)
++ total_pages: `20` (number)
```
psrpinto, line 66: Q: Should we make this optional? I think there are cases where we might not now how many total items there are, for example, recurring calendar events.

ketlane: The whole pagination block is optional (not included) by default. We will only include it for the endpoints where we count the total number of elements in the collection (currently nowhere, the first one will be project items if we agree on this form of pagination).

But if we do count the total number of elements, then all these 4 properties can be calculated.

psrpinto: What I meant is that in other data structures in this file we explicitly define which fields are required and which are optional. Should we not do the same here?

psrpinto: EDIT: I see we don't do it for all the data structures but maybe we should.

jenssegers (approved):

jdrieghe (approved):

adognameddog (approved):

JordyDeGroeve (approved):

src/datastructures.apib
ketlane: @regularjack I would like to discuss this on Slack, to better understand your concern. However, I will merge the PR for now since it has a majority already and if we do find, on Slack, that there needs to be something changed, I will create a follow-up PR.

PR #259, opened 2018-12-26
Change pagination response format

jenssegers:
After some more discussion, we thought this API response format was more in line with our pagination request options (page size/number).

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -87,8 +87,7 @@ Sample response with pagination:
     "meta": {
         "page": {
             "size": 20,
-            "number": 2,
-            "total": 10,
+            "number": 2
         },
         "matches": 200
     }
@@ -3388,7 +3387,6 @@ Unregister a webhook.
 ## Pagination
 + size: `10` (number)
 + number: `2` (number)
-+ total: `20` (number)
 
 ## PaymentTerm  (object)
 + type (enum[string])
--- a/apiary.apib
+++ b/apiary.apib
@@ -84,11 +84,13 @@ Sample response with pagination:
             "email": "john@piedpiper.eu"
         }
     ],
-    "pagination": {
-        "count": 200,
-        "per_page": 2,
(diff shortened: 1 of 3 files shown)


adognameddog (dismissed):

mrtus (approved):

apiary.apib
```diff
@@ -84,11 +84,12 @@ Sample response with pagination:
             "email": "john@piedpiper.eu"
         }
     ],
-    "pagination": {
-        "count": 200,
-        "per_page": 2,
-        "current_page": 1,
-        "total_pages": 100,
+    "meta": {
```
ketlane, line 87: Q: Should we build a `Meta` data structure, so that we can just include meta in individual endpoints? 

`+ meta (Meta)`


(Instead of having a repetitive block like the following in each endpoint that paginates)
`+ meta (object)`
`    + page (Pagination)`
`    + matches: 20 (number)`

mrtus: Wouldn't do that just yet, that would assume we would have pagination on every endpoint when we include other meta data.

ketlane: But we don't have any other meta data just yet...

ketlane (approved):

apiary.apib
stivni: go specific first, abstract or generalise later ;)

apiary.apib
```diff
@@ -84,11 +84,12 @@ Sample response with pagination:
             "email": "john@piedpiper.eu"
         }
     ],
-    "pagination": {
-        "count": 200,
-        "per_page": 2,
-        "current_page": 1,
-        "total_pages": 100,
```
stivni, line 91: fyi: we dropped this property for now, because we didn't agree on naming and is just a helper anyways
might add this in the future, if it makes more sense then

stivni (approved):

PR #260, opened 2018-12-28
[PRO-1285] Documentation for the update project item cost endpoint

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-1285

### What has been done
Added documentation for the update project item's cost endpoint


apiary.apib
```diff
@@ -2981,6 +2981,18 @@ Get a list of project items.
...
 
+### projectItems.update [POST /projectItems.update]
+
+Update a project item.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `cfb4146d-06be-41f1-bb39-aa3c929c71dc` (string, required)
+        + unit_internal_cost: `123.30` (number, optional)
```
adognameddog: Q: Shouldn't this be a money object? With an amount and a currency code?

ketlane: Yes, fixed.

adognameddog (approved):

adognameddog:
Added the hold label, since this functionality should not be made official just yet.

adognameddog (dismissed):

ArnaudWeyts (dismissed):

src/07-projects/project-items.apib
```diff
@@ -71,4 +71,16 @@ Get a list of project items.
                 + updated_at: `2018-12-05T16:44:33+00:00` (string)
         + meta (object)
             + page (Pagination)
-            + matches: `120` (number)
\ No newline at end of file
+            + matches: `120` (number)
+
+### projectItems.updateCost [POST /projectItems.updateCost]
```
jenssegers, line 115: Q: Why not `.update`?

ArnaudWeyts (dismissed):

src/07-projects/project-items.apib
adognameddog: A: Because we only allow to update the cost.

adognameddog (dismissed):

src/07-projects/project-items.apib
jenssegers: If that field is the only thing that makes sense updating for a project item, in general, I think it's still an option. Will there be an update endpoint in the future? And if so, what will it include besides the cost?

I understand the reasoning to go for an explicit endpoint, but mainly thinking about consistency here. If we plan to add an `update` method in the future we can also already release that but with only the cost field for now.

If updating the cost has a bunch of side-effects and can't be done in combination with other properties being updated, then let's go for this endpoint. Just want to make sure it checks the boxes for being a separate endpoint.

adognameddog (approved):

ArnaudWeyts (approved):

joaoinacio (approved):

nickwaelkens (approved):

src/07-projects/project-items.apib
ketlane: Updating the cost is quite specific as it also needs to update the purchase price of the originating order item, as a side-effect.

PR #261, opened 2018-12-31
Filter projects list on term [PRO-1182]

adognameddog:
Added a filter on `projects.list` to search on term. It searches project title and description.


adognameddog:
Hmmm, some commits do not belong here. Gonna try again.

PR #263, opened 2019-01-03
Add new types to migrate.id endpoint 

tiagobastos:
Matching the accepted types with the ones supported at the moment.

Removing supported types is a backward compatibility break.
But a warning is stated in the description of this end-point that that compatibility is not guaranteed.

An alphabetically sorting was done. Better to review commit by commit.

### DIFF
--- a/src/99-other/migrating.apib
+++ b/src/99-other/migrating.apib
@@ -25,10 +25,8 @@ Translates an ID from the deprecated API into a new UUID.
                 + milestone
                 + product
                 + project
-                + subscription
                 + task
                 + timeTracking
-                + ticket
                 + user
         + id: `1` (number)
 
--- a/src/99-other/migrating.apib
+++ b/src/99-other/migrating.apib
@@ -37,20 +37,20 @@ Translates an ID from the deprecated API into a new UUID.
             + type: `contact` (enum[string])
                 + Members
                     + account
-                    + user
-                    + product
-                    + contact
                     + company
+                    + contact
+                    + creditNote
+                    + customFieldDefinition
                     + deal
                     + dealPhase
-                    + project
-                    + milestone
-                    + todo
                     + event
-                    + ticket
                     + invoice
-                    + creditNote
-                    + subscription
+                    + milestone
+                    + product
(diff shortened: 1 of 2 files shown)


WouterSioen (approved):

src/99-other/migrating.apib
```diff
@@ -13,21 +13,21 @@ Translates an ID from the deprecated API into a new UUID.
         + type: `contact` (enum[string])
             + Members
                 + account
-                + user
-                + product
-                + contact
+                + call
                 + company
+                + contact
+                + creditNote
```
dietrichm, line 19: B: should be `credit_note`.

src/99-other/migrating.apib
```diff
@@ -13,21 +13,21 @@ Translates an ID from the deprecated API into a new UUID.
...
+                + customField
```
dietrichm, line 20: B: should be `custom_field`.

src/99-other/migrating.apib
```diff
@@ -13,21 +13,21 @@ Translates an ID from the deprecated API into a new UUID.
...
+                + product
+                + project
                 + task
-                + meeting
-                + call
-                + ticket
-                + invoice
-                + creditNote
-                + subscription
+                + timeTracking
```
dietrichm, line 29: B: should be `time_tracking`.

dietrichm (dismissed):
B: some accepted types are using underscores. Only `dealPhase` doesn't. Maybe we should fix this first.

tiagobastos:
@dietrichm I have a branch on core that will change the accepted types mentioned to camelCase. 
Will do the PR after the blocking as passed.

dietrichm:
@tiagobastos that will have to be implemented in a BC supporting way. 👍

luisaterrabrito (approved):

jenssegers (approved):

dietrichm (approved):

JordyDeGroeve (approved):

tiagobastos:
Discussed in a meeting. Decided for another approach.

PR #265, opened 2019-01-08
Return time tracking and timer subject's related objects

dietrichm:
To let API clients more easily process the related objects for a time tracking or running timer's subject, we propose to add the related objects' types and UUIDs to the return data.

For example, a time tracking can be tracked on a milestone, but that milestone will be linked to a project which in turn will be linked to a contact or company. Making a list of these related objects explicit in the return data is valuable for some clients and our own mobile app.

The new `relates_to` property contains an array of related entities returned in the `data` of `timeTracking.list`, `timeTracking.info` and `timers.current`. This property is not writeable and thus not available in `.add` or `.update` endpoints.

---

Best to review the whole diff at once.


WouterSioen:
I'd like to see feedback from the users of this endpoint too @Kemosabert @JorenSaey @mikeverf @braemJef @kevinwaelkens

WouterSioen (dismissed):
Looks good to me but I'd like to see feedback from the users from this endpoint.

JordyDeGroeve (dismissed):

kevinwaelkens:
I heard Charlotte talk about this, but I don't think if this has any use to my timetrackings. I think it's just best to show the first level of related items, not anything deeper?

JorenSaey:
It makes it easier for milestone specific. You don't need to wait for request to finish in order to fetch project and customer. Still this data will need to fetched in sagas so it takes away a bit of the complexity for us. Makes it easier to potentially side load the entities

JorenSaey (dismissed):

tiagobastos (dismissed):

mikeverf (dismissed):

paultimoce (dismissed):

JordyDeGroeve (dismissed):

tiagobastos (dismissed):

mikeverf (dismissed):

WouterSioen (dismissed):

jenssegers (dismissed):

apiary.apib
```diff
@@ -3049,6 +3049,18 @@ Get a list of tracked time.
                             + ticket
                     + id: `58f5b799-51c4-4eb9-8308-b1aa02e0a873` (string)
                 + invoiceable: true (boolean)
+                + relates_to (array)
```
jenssegers, line 3052: My eye still slightly twitches reading this. Should this be similar to the link(s) we had on events? Just not sure if this is the right "term" in a time tracking context. But could also just be me.

JorenSaey (dismissed):

psrpinto:
Not sure I understand the use case here. For the example given, the `milestone` that belongs to a `project`, which in turn is linked to a `contact` and/or `company`:

If the API consumer needs to get the related entities of multiple `timetracking`s, wouldn't it be a matter of querying `/projects.list` with a `milestone_ids` filter, and use sideloading for the `contact` and/or `company`?

Feels like this use case could be accomplished by using other building blocks we already have in place.

WouterSioen:
@regularjack theoretically, everything in this PR is already possible. There's 7 possible different cases though (and that's not including edge cases), which would make this really complex to implement in a front-end (or even multiple front-ends).

We first considered moving the logic to the front-end, but it made sense to add this convenience in the API by default instead of expecting all integrations that need this to build this themselves.

apiary.apib
dietrichm: Links on calendar events are updateable explicitly. In this case they are calculated (on retrieval or saving, that needs to be decided) from the subject of the time tracking. So it's not exactly the same.

dietrichm:
Not meant to be merged as a whole.

apiary.apib
```diff
@@ -3049,6 +3049,18 @@ Get a list of tracked time.
...
+                    + (object)
+                        + type: `company` (enum[string])
+                            + Members
+                                + company
+                                + contact
+                                + event
+                                + todo
```
psrpinto: Should be `task`.

psrpinto (dismissed):
Approving but `todo` is now known as `task`

psrpinto (approved):

WouterSioen (approved):

JordyDeGroeve (approved):

JordyDeGroeve:
Won't this give any problems when trying to sideload tasks?

~PS: conflicts~ 😞 
not meant to be merged

ketlane (approved):

braemJef (approved):

Kemosabert (approved):

apiary.apib
jenssegers: `tracked_on`?

tiagobastos (approved):

jenssegers (approved):

kevinwaelkens (approved):

mikeverf (approved):

JorenSaey (approved):

WouterSioen:
Can we have a final check from @stivni or an alternative suggestion?

stivni (approved):
Seems fine to me!

@regularjack I think this approach also makes more sense if you want to query on all timetrackings that "relate to" a certain customer for a certain period... (if we would provide that as a filter) regardless of it was tracked on a project or tasks or ... It also makes the sideloading easier and in one hop.

stivni:
I am however interested in how you are going to implement this :)

dietrichm:
To be split for time trackings and timers.

PR #266, opened 2019-01-11
Replace `rate` with `id` and `type` in `tax` object on `invoice.info`, `creditNote.info` and `quotations.info` responses

dcsg:
Bookkeeping integrations need to have more details regarding the Tax Rate. The percentage is not enough since the same percentage value can have different meanings. By exposing the Tax Rate ID we give them the possibility to fetch more details of that given Tax Rate.

This change implies the removal of the `rate` from the `tax` object in `invoices.info`, `creditNote.info` and `quotations.info` response. In exchange, we are adding the `id` and the `type` to that object. That way Integrators will be able to fetch the details of the Tax Rate.


src/05-invoicing/invoices.apib
```diff
@@ -141,6 +141,7 @@ Get details for a single invoice.
                                         + excluding
                             + tax (object)
                                 + rate: 0.21 (number)
```
dcsg: Q: since we are now exposing the Tax Rate ID, should we remove this and only send the ID?

luisaterrabrito: I vote on only sending the UUID like in the draft tax_rate_id

apiary.apib
```diff
@@ -2067,6 +2067,7 @@ Get details for a single invoice.
                                         + excluding
                             + tax (object)
                                 + rate: 0.21 (number)
+                                + id: `e2314517-3cab-4aa9-8471-450e73449042` (string)
```
mrtus, line 2133: Q: Now that we are providing an ID we also need a `type: taxRate` right?

dcsg: A: I think that is not necessary, but if this creates some confusion I can rename it to `rate_id`?

src/05-invoicing/invoices.apib
lowiebenoot: I think it would make more sense indeed, if we replace tax with taxRate and only provide the id.
Something like this:


```
+ tax_rate (object)
    + type: `tax_rate` (string)
    + id: `4b0f959a-27e3-4080-b1b2-a05b374863cb` (string)
```

On the other hand, there currently is no `taxRates.info` endpoint. So if you want to know the rate or the name, you need to fetch them via the `taxRates.list` endpoint, which is a bit weird.

apiary.apib
mrtus: But that is our format right? We should keep it consistent.

It's also done for `witheld_tax` so why shouldn't we do it for `tax`?

src/05-invoicing/invoices.apib
luisaterrabrito: The taxRates.list doesn't allow to filter on the taxRate id. @PieterReel should we then add a taxRates.info or a an if filter on the list?

dcsg: @luisabritotl IMO both options are valid.

@lowiebenoot why do you need the `type`? and why not just:
```
+ tax_rate_id: `4b0f959a-27e3-4080-b1b2-a05b374863cb` (string)
```

luisaterrabrito: I think Daniel that is a matter of consistency

dcsg: Indeed,  I just realized that we add a `type` for almost "all" objects.

apiary.apib
dcsg: Indeed! Thanks for pointing out.

lowiebenoot (dismissed):

jdrieghe (approved):

jdrieghe (dismissed):
Shouldn't the same changes be applied to `/creditNotes.info` and `/quotations.info`?

JordyDeGroeve (commented):
Q: isn't this a change that should be documented in the changelog?

dcsg:
> Shouldn't the same changes be applied to /creditNotes.info and /quotations.info?

Only in `creditNotes.info` and already added.

luisaterrabrito (approved):

jdrieghe (approved):

EOSullivanBerlin (approved):

lowiebenoot (approved):

nielsvandenb:
@dcsg can we ship this? Partners are waiting for this :)

dcsg:
@nielsvandenb this is waiting for a new API Version release. Not sure when that will happen. Better to ask @stivni or @PieterReel .

Matthijs-Huisman:
> 
> 
> > Shouldn't the same changes be applied to /creditNotes.info and /quotations.info?
> 
> Only in `creditNotes.info` and already added.

Why not in Quotations? Would help me a lot.

jenssegers (approved):

stivni:
> > > Shouldn't the same changes be applied to /creditNotes.info and /quotations.info?
> > 
> > 
> > Only in `creditNotes.info` and already added.
> 
> Why not in Quotations? Would help me a lot.

Hi @menonweb!

This will be adapted in the future, but out of the scope of this change.
We will list your request with the other we already received, which integrator or customer are you asking this for?

stivni:
@dcsg can rebase to `2019-03-13`?
And can you also provide an entry in the changes document?

In that case, this can actually be merged to that branch

Matthijs-Huisman:
> 
> 
> > > > Shouldn't the same changes be applied to /creditNotes.info and /quotations.info?
> > > 
> > > 
> > > Only in `creditNotes.info` and already added.
> > 
> > 
> > Why not in Quotations? Would help me a lot.
> 
> Hi @menonweb!
> 
> This will be adapted in the future, but out of the scope of this change.
> We will list your request with the other we already received, which integrator or customer are you asking this for?

Hi @stivni, It's a connection for Databrydge between Teamleader & Exact Online for a client, can't disclose publicly who it is.

jdrieghe (approved):

luisaterrabrito (approved):

lowiebenoot (approved):

src/05-invoicing/invoices.apib
```diff
@@ -143,7 +143,8 @@ Get details for a single invoice.
                                     + Members
                                         + excluding
                             + tax (object)
```
lowiebenoot, line 145: Would it make sense if we also rename the property from `tax` to `taxRate`?

steftrenson (approved):

PR #268, opened 2019-01-15
User invites endpoint

mrtus:
In order to improve our user creation flow, we should create extra API endpoints that are able to invite a person using an email address, cancel an invite and list the invites.

### Note
- I'll build the file once we are ready to publish this.

### Added
- An endpoint that will allow users to be invited using an email address.
- An endpoint that will allow an invite to be cancelled.
- An endpoint that will allow all invites to be listed.


apiary.apib
```diff
@@ -536,6 +536,17 @@ Get details for a single user.
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
 
+### users.invite [GET /users.invite]
```
rathesDot: This should be a post request, right?

apiary.apib
```diff
@@ -536,6 +536,17 @@ Get details for a single user.
...
+
+Invite a single user to your account
+
++ Request (application/json)
+
+    + Attributes (object)
+        + email: `john@teamleader.eu` (string, required)
```
rathesDot, line 903: Should this also contain an `account_id`?

rathesDot (dismissed):

apiary.apib
mrtus: 🤦‍♂️ It should

apiary.apib
mrtus: Any reason? The user that invites a new user is already authenticated so, we know for which account this user is invited.

src/01-general/users.apib
```diff
@@ -117,3 +117,14 @@ Get details for a single user.
                     + tr-TR
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
+
+### users.invite [GET /users.invite]
```
mrtus:
```suggestion
### users.invite [POST /users.invite]
```

jenssegers (dismissed):

rathesDot (dismissed):

src/01-general/invites.apib
```diff
@@ -0,0 +1,34 @@
...
+            + email: `john@teamleader.eu` (string, optional)
+
++ Response 200 (application/json)
+    + Attributes
+        + data (object)
+            + id: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5` (string)
+            + first_name: `john` (string)
+            + last_name: `doe` (string)
+            + email: `john@teamleader.eu` (string)
+            + created_at: `2017-05-09T11:25:11+00:00` (string)
```
dcsg: Q: do we need the time here or the date is just enough?

src/01-general/invites.apib
```diff
@@ -0,0 +1,34 @@
+## Invites [/invites]
+
+*Required scopes: `users`*
+
+### invites.list [GET /invites.list]
```
dcsg: Q: Should we make this endpoint more specific? IMHO `invite` is quite generic and since this is coupled to the user creation flow I would even suggest something link `/users.invites` for the list and `/users.cancelInvite` to cancel a specific invite? Thoughts?

src/01-general/invites.apib
mrtus: That would be up to the front-end to decide. How we have it in the API response doesn't really matter, the more detailed the better I'd say.

src/01-general/invites.apib
mrtus: An invite is not a `/users` resource and thus has it's own resource endpoint `/invites` An invite is not a user in any way.

dcsg: Indeed, but the invite is for a user, would `/user-invites.list` make more sense then? Because only `/invite` is too generic IMHO.

src/01-general/invites.apib
mrtus: We'll have to call it `invited_at` though.

src/01-general/invites.apib
stivni: I'm actually against `users.invites` (because I want to avoid subresources), but no strong opinion on `invites` vs `userInvites` (mind the camelCasing).

jenssegers: No strong opinion here as well. `userInvites` would indeed make it more explicit what kind of invite we're talking about. Although I don't think we don't have any other invitation system? Then again, once we start working with inviting remote people to calendar events this might become a bit more confusing. So in favor of `userInvites` :ok_hand:

dcsg: I believe we would be safer with `userInvites` as well.

rathesDot: I don't see the benefit of calling it `userInvite` ... If you start adding remote people for the calendar, then you are also inviting users there. So it won't be better in that case, right?

I personally would stick with invites here and then go for a more accurate solution for inviting people to events when we actually start doing it.

(An option, for example, would be to use the same endpoint for both because the logic on the API level will be the same)

jenssegers: I was also going to suggest going for this name for now. But because we already saw that it could cause conflicts in other domains of our application I suggest to rename it now rather than having to introduce a BC breaking change later. There's no downside to renaming it, only making it more explicit.

rathesDot: But where is userInvites more explicit that just invites? `userInvites` could also be the invites endpoint for `inviting users to events`, right?

jenssegers: Inviting someone to an event is a different action with different side effects and maybe even other required properties for sending the invite. So I would not mix them together, that would only make things very confusing for third party developers.

rathesDot: I was saying that `userInvites` could also be understood as the endpoint for `inviting users to events` ... So that it is not clear either ...

So if we say invites is not clear enough, then `userInvites` isn't either... We should then think about a more clear name

mrtus: Should we maybe take this discussion offline? There are some different opinions here and commenting back and forth won't really help us here I'd say.

dcsg: We discussed this offline and the decision was to make it `/userInvites.*`

dcsg (dismissed):

rathesDot (dismissed):

apiary.apib
```diff
@@ -625,6 +625,54 @@ Get details for a single user.
...
+        + last_name: `doe` (string, required)
+        + email: `john@teamleader.eu` (string, required)
+
++ Response 204
+
+## Invites [/userInvites]
+
+*Required scopes: `users`*
+
+### invites.list [GET /userInvites.list]
```
jenssegers: These titles should also be renamed right?

```suggestion
### userInvites.list [GET /userInvites.list]
```

dcsg: Indeed, I also changed other titles/phrases.

dcsg (dismissed):

dcsg (dismissed):

jenssegers (approved):

rathesDot (dismissed):

apiary.apib
```diff
@@ -664,6 +670,54 @@ Get details for a single user.
...
+Invite a single user to your account
+
++ Request (application/json)
+
+    + Attributes (object)
+        + first_name: `john` (string, required)
+        + last_name: `doe` (string, required)
+        + email: `john@teamleader.eu` (string, required)
+
++ Response 204
```
jenssegers: We should return a `201` with the invitation id.

jenssegers (dismissed):

dcsg (dismissed):

dcsg (dismissed):

rathesDot (dismissed):

dcsg (dismissed):

rathesDot (dismissed):

jenssegers (dismissed):

apiary.apib
```diff
@@ -792,6 +801,75 @@ Get details for a single user.
                     + active
                     + deactivated
 
+### users.invite [POST /users.invite]
+
+Invite a single user to your account
```
jenssegers, line 896:
```suggestion
Send out an invite for a user to join your account.
```

apiary.apib
```diff
@@ -792,6 +801,75 @@ Get details for a single user.
...
+            + id: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5` (string)
+            + first_name: `john` (string)
+            + last_name: `doe` (string)
+            + email: `john@teamleader.eu` (string)
+            + invited_at: `2017-05-09T11:25:11+00:00` (string)
+            + expires_at: `2017-05-12T11:25:11+00:00` (string)
+
+### userInvites.cancel [GET /userInvites.cancel]
+
+Cancel a user invite.
```
```suggestion
Cancel a user invitation.
```

jenssegers (dismissed):

jenssegers (dismissed):

PR #272, opened 2019-01-21
Add proposal for the users.list filter on term

WouterSioen:
Add a filter on the users.list endpoint.

It will filter on first name, last name and email.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -466,7 +466,7 @@ Get a list of all users.
     + Attributes (object)
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
-            + term: `John` (string, optional) - Filters on username, real name and email address
+            + term: `John` (string, optional) - Filters on username, first_name, last_name and email
         + sort (array, optional)
             + (object)
                 + field (enum[string], required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -466,7 +466,7 @@ Get a list of all users.
     + Attributes (object)
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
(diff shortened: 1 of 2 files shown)


marcwestermann:
Hey Wouter, thanks for picking this up! The name (first name and last name) is the most interesting filter for us. We are trying to filter the users on an API level to select them as attendees on the new calendar. Maybe, function (position) would also be nice to have, but not required.
image
@duivvv , what do you think? :)

geoffreydhuyvetters:
I think function is nice to have but not essential.
The combination `company` / `function` makes sense in the context of a contact (> 'ceo madewithlove' for example)

src/01-general/users.apib
```diff
@@ -48,6 +48,7 @@ Get a list of all users.
     + Attributes (object)
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
+            + term: `John` (string, optional) - Filters on username, real name and email address
```
dietrichm: B: there is no notion of "real name" in the return data for users. The first and last name is returned as a result from splitting on the first space.

```suggestion
            + term: `John` (string, optional) - Filters on username, first_name, last_name and email
```

Also the username still needs to be added to the user info.

dietrichm (changes requested):

dietrichm (dismissed):
PP: username still should be added to the return data.

mrtus (dismissed):

apiary.apib
```diff
@@ -466,6 +466,7 @@ Get a list of all users.
     + Attributes (object)
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
+            + term: `John` (string, optional) - Filters on username, first_name, last_name and email
```
mrtus: Q: Should we leave out username for now? I don't think it has any use for now.

dietrichm (approved):

JordyDeGroeve (approved):

mrtus (approved):

marcwestermann (approved):

tiagobastos (approved):

jenssegers (approved):

PR #274, opened 2019-01-24
2019-01-24

stivni:
This are the breaking changes that will go live today.

fyi : the actual api changes were already approved, but the changelog changed


apiary.apib
```diff
@@ -327,6 +327,27 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and will not appear in this list.
 
+#### Beta
```
tiagobastos: We should update this with today's date.
```suggestion
#### 2019-01-24
```

apiary.apib
```diff
@@ -327,6 +327,27 @@ _This will only affect the version for those API calls and won't affect any othe
...
+
+- The property `local_business_number` was renamed to `national_identification_number` on the following endpoints:  
+  `companies.info`,  
+  `companies.list`,  
+  `companies.add`,  
+  `companies.update`,  
+  `invoices.info`  
+  and `creditNotes.info`.
+
+- Added the bookyear name to the `invoice_number` on `invoices.list` and `invoices.info`.
```
jdrieghe: I order to be consistent with language, we should use passive voice here ("bookyear name was added") or use active voice in the previous bullet point "Renamed x to x in the following endpoints".

jdrieghe (dismissed):

tiagobastos (dismissed):

apiary.apib
```diff
@@ -344,9 +344,8 @@ We list all backwards-incompatible changes here. As described above, new additio
   `invoices.info`  
   and `creditNotes.info`.
 
-- Added the bookyear name to the `invoice_number` on `invoices.list` and `invoices.info`.
-  Earlier it was only returning the number, e.g. `"27"`.
-  Now it adds the name of the bookyear as well, e.g. `"2018 / 27`".
+- Invoice number actually contain the bookyear, so from now on, you will see eg. `2018 / 27` instead of `27`.
```
jdrieghe: Invoice numberS (plural here) or use "contains"

apiary.apib
stivni: better in line now?

apiary.apib
stivni: solved it differently : `An invoice number actually contains the bookyear, ...`

tiagobastos (dismissed):

jdrieghe (dismissed):

luisaterrabrito (dismissed):

src/changes.apib
```diff
@@ -48,6 +48,26 @@ _This will only affect the version for those API calls and won't affect any othe
...
+  Therefore the `status` property in `invoices.list` and `invoices.info` will either return `outstanding` or `matched` for invoices that were previously just `booked`.  
+  This is also reflected in the `status` filter on `invoices.list`: to allow you to still query all `outstanding` and `matched` invoices (previously all `booked`) , it now *only* accepts an array of filter values (eg. `["outstanding", "matched"]` or `["draft"]`).
+
+- The property `local_business_number` was renamed to `national_identification_number` on the following endpoints:  
+  `companies.info`,  
+  `companies.list`,  
+  `companies.add`,  
+  `companies.update`,  
+  `invoices.info`  
+  and `creditNotes.info`.
```
dietrichm: B: this is really showing on distinct lines in the generated HTML. Can we put these on a single line?

dietrichm (dismissed):

src/changes.apib
stivni: seems rather PP to me 🤔

stivni: tbh, I actually want readers to spot endpoint names easily in the changelogs, so I would like to try it like this...

dietrichm: Can we use list items in that case? Because it looks like faulty HTML or an unintended result right now.

jdrieghe (approved):

JordyDeGroeve (approved):

luisaterrabrito (approved):

dietrichm (approved):

tiagobastos (approved):

PR #282, opened 2019-02-04
Contacts National Identification Number

tiagobastos:
### Added
* National Identification Number to contacts endpois:
	* `/contacts.list`
	* `/contacts.info`
	* `/contacts.add`
	* `/contacts.update`


src/02-crm/contacts.apib
```diff
@@ -73,6 +73,7 @@ Get a list of contacts.
                 + added_at: `2016-02-04T16:44:33+00:00` (string)
                 + updated_at: `2016-02-05T16:44:33+00:00` (string)
                 + `web_url`: `https://app.teamleader.eu/contact_detail.php?id=2a39e420-3ba3-4384-8024-fa702ef99c9f` (string)
+                + `national_identification_number`: BE 4283 (string)
```
WouterSioen: Not sure if the example is a good one, since I have no idea what a national identification number looks like, but this might be slightly confusing with the taxcode here.

WouterSioen (dismissed):

src/02-crm/contacts.apib
dietrichm: We could pick one in a format mentioned on https://en.wikipedia.org/wiki/National_identification_number#Europe

I.e. in Spain it's in the format `00000000-A`, where _0 is a digit and A is a checksum letter_.

JordyDeGroeve (dismissed):
Would go for the format that Dietrich proposed.

src/02-crm/contacts.apib
dietrichm: PP: it would be better to move this property somewhere after IBAN and BIC, before the more "preferency" properties. `web_url` is a metadata type of property so I wouldn't return more user provided entity properties after that one.

src/02-crm/contacts.apib
```diff
@@ -135,6 +135,7 @@ Get details for a single contact.
             + added_at: `2016-02-04T16:44:33+00:00` (string)
             + updated_at: `2016-02-05T16:44:33+00:00` (string)
             + `web_url`: `https://app.teamleader.eu/contact_detail.php?id=cde0bc5f-8602-4e12-b5d3-f03436b54c0d` (string)
+            + `national_identification_number`: BE 9823 (string)
```
dietrichm: PP: it would be better to move this property somewhere after IBAN and BIC, before the more "preferency" properties. `web_url` is a metadata type of property so I wouldn't return more user provided entity properties after that one.

src/02-crm/contacts.apib
```diff
@@ -177,6 +177,7 @@ Add a new contact.
         + tags: prospect, expo (array[string], optional)
         + custom_fields (array[CustomFieldValue], optional)
         + marketing_mails_consent: false (boolean, optional)
+        + `national_identification_number`: BE 3478 (string)
```
dietrichm: PP: it would be better to move this property somewhere after IBAN and BIC, before the more "preferency" or metadata properties.

src/02-crm/contacts.apib
```diff
@@ -227,6 +227,7 @@ Update a contact.
         + tags: prospect, expo (array[string], optional)
         + custom_fields (array[CustomFieldValue], optional)
         + marketing_mails_consent: false (boolean, optional)
+        + `national_identification_number`: BE 1290 (string)
```
dietrichm: PP: it would be better to move this property somewhere after IBAN and BIC, before the more "preferency" or metadata properties.

src/02-crm/contacts.apib
tiagobastos: Q: Would this also mean changing the order its returned in the actual response too?

WouterSioen (dismissed):

JordyDeGroeve (dismissed):

rathesDot (dismissed):

src/02-crm/contacts.apib
dietrichm: Yes preferably.

dietrichm (dismissed):

src/02-crm/contacts.apib
tiagobastos: Added default value for `national_identification_number` under the field `bic`

WouterSioen: FYI: returned order does only matter in the documentation

Json does not necessarily keep it's order (f.e. Chrome reorders json objects alphabetically)

apiary.apib
```diff
@@ -775,6 +775,7 @@ Get a list of contacts.
                 + birthdate: `1987-04-25` (string)
                 + iban: `BE12123412341234` (string)
                 + bic: `BICBANK` (string)
+                + `national_identification_number`: 86792345-L (string)
```
jenssegers: I don't think the key should have the `s but the value 😬

jenssegers (dismissed):
Blocking because I'm not sure what apiary will do when the key is wrapped with `

apiary.apib
tiagobastos: 🙈

tiagobastos:
Sorry for this mistake guys. 🙈

WouterSioen (approved):

JordyDeGroeve (approved):

rathesDot (approved):

dietrichm (approved):

PR #289, opened 2019-02-06
Return the invoiceable user preference

WouterSioen:
Based on https://github.com/teamleadercrm/api/pull/231, we're going to return the `invoiceable` property of a user as a preference in users.me

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -476,6 +476,8 @@ Get the current authenticated user.
                     + tr-TR
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
+            + preferences (object)
+                + invoiceable: true (boolean)
 
 ### users.list [GET /users.list]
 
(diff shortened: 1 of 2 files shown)


JordyDeGroeve (approved):

apiary.apib
```diff
@@ -476,6 +476,8 @@ Get the current authenticated user.
                     + tr-TR
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
+            + preferences (object)
+                + invoiceable: true (boolean)
```
JordyDeGroeve, line 480: Q: should we make it clear what is invoiceable?
And maybe add that this is the default value?

dietrichm (approved):

apiary.apib
WouterSioen: Not sure if this is necessary. I think it's pretty clear, right?

WouterSioen: If it's not clear, we should probably give it a better name.

tiagobastos (approved):

kevinwaelkens (approved):

rathesDot (approved):

PR #290, opened 2019-02-08
Proposal: WorkOrder endpoints

dietrichm:
To be able to rebuild the WorkOrder functionality into the new mobile application, we are designing and building new API endpoints to manage the WorkOrders and related data.

A few major distinctions compared to the endpoint in v1 are that we introduce draft WorkOrders, separate endpoints to manage images (usually photos included in the WorkOrder PDF) and make explicit the various actions and some of the side effects (i.e. signing, finalising and sending).

---

The goal of this PR is to discuss WorkOrder API design, **not** to merge this as a whole. We will be gradually implementing and documenting parts (endpoints and/or a subset of provided properties) of this API proposal over several iterations.

Best to review the complete diff at once.


JorenSaey (dismissed):
It seems that it has everything needed to complete the first flow for adding a draft

tiagobastos (dismissed):

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
++ intervention_address (object, optional)
+    + One Of
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + address (Address)
++ distance: `25.2` (number, optional)
```
JordyDeGroeve: Q: should we pass a unit like we do with currency?
PP: at least add a comment to specify this is in km.

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
++ event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
++ intervention_address (object, optional)
+    + One Of
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + address (Address)
```
JordyDeGroeve: Feels strange to call it `intervention_address` and pass a customer. What was the reasoning behind this?

WouterSioen: There are two ways you can show that the intervention happened at a different address than the one of the customer:

1. it is the address of a second customer (f.e. an overlapping company needs to get the bill for work done on one of their sub-companies)
2. it is a hand-filled address (so not coupled to a customer)

WouterSioen: `Intervention location` might make more sense than `intervention_address` in that case? (still with a customer or an address)

JordyDeGroeve: `intervention location` seems better to me.

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
         + timeTracking.deleted
         + timeTracking.updated
         + user.deactivated
+
+## WorkOrderData (object)
```
stivni: PP: I actually don't like that complete responses are modelled as datastructures
there is not a lot of reusability in that and it makes you have to check two files all the time
I used datastructures for things that are value-object like things

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
++ customer (object, required)
+    + type (enum[string], required)
+        + Members
+            + contact
+            + company
+    + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
++ datetime: `2017-05-26T10:01:49+00:00` (string, required)
```
stivni: Q : what does this `datetime` represent?

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
++ description (string, optional)
```
stivni: Q : is this rich text (markdown?)?
should we mention?

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string)
+        + Include WorkOrderData
+
++ Response 204
+
+### workOrders.drafts [GET /workOrders.drafts]
```
stivni: PP : I don't think we need a separate endpoint for that

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
++ department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional)
++ layout_id: `a05320f6-b3ef-4ff9-a287-2074a12fb3e9` (string, optional)
++ milestone_id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string, optional)
++ event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
```
stivni: Q : why do we currently need the event id?

src/datastructures.apib
stivni: but a contact or a customer can have different addresses... so shouldn't we provide more detail that just the link? and what if the address changes over time? I think storing the address in all cases, makes more sense... Keeping a link to where the address comes from, might make sense

src/datastructures.apib
stivni: I like the idea about the units, although I don't think we'll go into imperial stuff soon 🙈
Not sure if it's worth it to always have to provide that, although I like the explicitness of it

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
+    + One Of
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + address (Address)
++ distance: `25.2` (number, optional)
++ users (array, required)
```
stivni: Q : are these the "users" of the work order?

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
+    + (object)
+        + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+        + hourly_rate: `65` (number, optional)
```
stivni: B : a rate is money, so should be amount & currency

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
+        + work_hours (array, optional)
+            + (object)
+                + One Of
```
stivni: Q : do we need to provide both ways?

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
+                    + Properties
+                        + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                        + duration: 3600 (number, required) - In seconds
+                    + Properties
+                        + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                        + ended_at: `2017-04-26T16:02:00+00:00` (string, required)
+                + description (string, optional)
++ line_items (array, optional)
```
stivni: Q: is this really "line items"?
Aren't these the "used products" or something?

src/datastructures.apib
```diff
@@ -140,3 +140,121 @@
...
+    + (object)
+        + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
+        + quantity: `3` (number, optional)
+        + description: `An awesome product` (string, optional)
+        + unit_price (object, optional)
```
stivni: Q : I'm not sure if this API should care about prices?

stivni: Same actually for descriptions etc

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
+
+### workOrders.update [POST /workOrders.update]
+
+Update a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string)
+        + Include WorkOrderData
```
stivni: PP : I'm not sure if we should allow updating all info on the workorder

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+        + image_data: `data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAIAAACQkWg2AAAAAXNSR0IArs4c6QAAAVlpVFh0WE1MOmNvbS5hZG9iZS54bXAAAAAAADx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8AAAASUVORK5CYII=` (string, required) - PNG image, base64 encoded
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (object)
```
stivni, line 226: Q: Why would we return an id for that image? Is there such a thing as a workorderImage?
Shouldn't a list of image be part of the workorder detail?

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
+            + type: `workOrderImage` (string)
+            + id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string)
+
+### workOrders.listImages [GET /workOrders.listImages]
```
stivni: PP : see previous comment, should be part of the details imo

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
+
+### workOrders.sign [POST /workOrders.sign]
+
+Sign a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+        + signature_data: `data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAIAAACQkWg2AAAAAXNSR0IArs4c6QAAAVlpVFh0WE1MOmNvbS5hZG9iZS54bXAAAAAAADx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8AAAASUVORK5CYII=` (string, required) - PNG image, base64 encoded
```
stivni: PP: `signature`

src/01-general/work-orders.apib
```diff
@@ -0,0 +1,199 @@
...
+
++ Response 204
+
+### workOrders.send [POST /workOrders.send]
+
+Send a finalized work order to the customer.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
```
stivni, line 290: Q : maybe we have to provide sending details here too? who to send it to, where to send duplicates, which layout to use, ...

stivni (dismissed):
Wow! Good job!

I have however a lot of remarks, stuff to discuss.

A more global main concern is, that we probably won't be able to publish this completely as described here.

I would recommend to make smaller PRs and deliver this step by step, in sync with what the mobile app wants to deliver.

stivni:
@dietrichm I only noticed the not about not to merge now, so ignore that part of my comment

src/01-general/work-orders.apib
stivni: but in that case we will need to have a "state" field on the workorder and a filter for it

stivni: I just assume we will have the need of showing combined lists of draft and finalized work orders...

apiary.apib
```diff
@@ -3250,6 +3250,208 @@ Update the current timer. Only possible if there is a timer running.
...
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string)
+        + Include WorkOrderData
+
++ Response 204
+
+### workOrders.drafts [GET /workOrders.drafts]
```
jenssegers: Q: Do we need a split between draft and finalized work orders?

src/datastructures.apib
dietrichm: Kept the `intervention_address` property but made the customer a property of the address, so we always have actual and non changing address fields but with an optional link to a customer entity.

apiary.apib
```diff
@@ -3519,3 +3721,121 @@ Unregister a webhook.
...
         + user.deactivated
+
+## WorkOrderData (object)
++ customer (object, required)
+    + type (enum[string], required)
+        + Members
+            + contact
+            + company
+    + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
++ datetime: `2017-05-26T10:01:49+00:00` (string, required)
```
jenssegers: Should be something that ends in `_at`

apiary.apib
```diff
@@ -3519,3 +3721,121 @@ Unregister a webhook.
...
++ description (string, optional)
++ department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional)
++ layout_id: `a05320f6-b3ef-4ff9-a287-2074a12fb3e9` (string, optional)
++ milestone_id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string, optional)
++ event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
++ intervention_address (object, optional)
```
jenssegers: Location?

apiary.apib
```diff
@@ -3519,3 +3721,121 @@ Unregister a webhook.
...
+    + One Of
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + address (Address)
++ distance: `25.2` (number, optional)
```
jenssegers: Distance of what to where? Driving distance?

apiary.apib
```diff
@@ -3519,3 +3721,121 @@ Unregister a webhook.
...
++ users (array, required)
```
jenssegers: Would try to find a fitting name inside the workorder domain that represents the people that will work or worked on this. Wikipedia seems to mention something like assignees.

apiary.apib
```diff
@@ -3519,3 +3721,121 @@ Unregister a webhook.
...
+            + (object)
+                + One Of
+                    + Properties
+                        + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                        + duration: 3600 (number, required) - In seconds
+                    + Properties
+                        + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                        + ended_at: `2017-04-26T16:02:00+00:00` (string, required)
+                + description (string, optional)
++ line_items (array, optional)
```
jenssegers: Does this represent the jobs to be performed?

src/datastructures.apib
dietrichm: For now this is plain text. I added an example description.

src/01-general/work-orders.apib
jenssegers: Following Stijn on this, not the biggest fan of the embedded data structures from the start too 🙈

src/01-general/work-orders.apib
dietrichm: To be discussed in meeting.

src/datastructures.apib
dietrichm: Providing a meeting ID will mark that meeting as done when the WorkOrder is finished.

src/datastructures.apib
dietrichm: These are the employees carrying out the work and tracking time. It might make sense to rename this to `employees` though.

Update: renamed to assignees as per @jenssegers' suggestion.

src/datastructures.apib
dietrichm: @JorenSaey what do you prefer here? Do you use a reusable component for creating and editing time trackings that supports both formats, or does one format suffice?

src/datastructures.apib
dietrichm: Correct. Don't recall why we went for line items anymore though. Likely to keep this in sync with how this is called in the invoicing domain.

src/datastructures.apib
dietrichm: It's possible a custom price is being used for an existing product. E.g. when a contractor is getting the product on short notice from a different supplier.

src/01-general/work-orders.apib
dietrichm: Since we have draft WorkOrders that can be very minimal it does make sense to allow updating most properties on an update, perhaps except the required ones such as customer.

Regarding the embedded data structures, I inlined those on my local branch since it's indeed a hassle when properties (sets) are different on different endpoints.

src/01-general/work-orders.apib
dietrichm: Agreeing on the latter, but we do need IDs to be able to delete images from draft WorkOrders.

src/datastructures.apib
dietrichm: Let's add a default fixed unit.

src/01-general/work-orders.apib
dietrichm: Distinct endpoint for drafts will go and a state property + filter on `.list` will be added.

src/datastructures.apib
dietrichm: To be checked with product for a suitable name.

src/datastructures.apib
dietrichm: Agreed to remove this from the proposal for now. Will be revisited later.

src/datastructures.apib
JorenSaey: I prefer just the sarted_at and duration combo, that is what we use in our TimeTracking module. :)

src/01-general/work-orders.apib
dietrichm: Additional properties will become obvious when we groom the relevant stories. Going to leave it like this for now.

apiary.apib
JordyDeGroeve: Yes but mostly products, I assume.
Something clearer than `line_items` would be nice though

apiary.apib
dietrichm: Fixed.

apiary.apib
dietrichm: See https://github.com/teamleadercrm/api/pull/290#discussion_r255568364

apiary.apib
dietrichm: Renamed to `products`.

src/datastructures.apib
dietrichm: It's the date when the work order was performed. I've renamed it to `performed_at`. We will check the naming with product anyway.

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+    + Attributes (object)
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + performed_at: `2017-05-26T10:01:49+00:00` (string, required)
+        + description: `Pipework intervention` (string, optional)
+        + department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional)
```
WouterSioen, line 3303: Q: is this really optional?

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+
+    + Attributes (object)
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + performed_at: `2017-05-26T10:01:49+00:00` (string, required)
+        + description: `Pipework intervention` (string, optional)
```
WouterSioen, line 3302: Q: can we keep this optional when we're starting to build an overview of workorders somewhere? I guess we'll need this to show which workorder is which?

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+        + department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional)
+        + layout_id: `a05320f6-b3ef-4ff9-a287-2074a12fb3e9` (string, optional)
+        + milestone_id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string, optional)
+        + event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
+        + intervention_address (object, optional)
+            + Include Address
```
WouterSioen: This should also be optional I guess? Can't we use an `OR` here?

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+                + type (enum[string], required)
+                    + Members
+                        + contact
+                        + company
+                + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + distance_travelled (object, optional)
+            + type: `km` (enum[string], required)
+                + Members
+                    + km
+            + value: `25.2` (number, required)
```
WouterSioen: Is this the total distance travelled? Or is this the average per assignee?

WouterSioen (dismissed):

apiary.apib
dietrichm: It is currently optional in the client API. We can fall back to the default department or leave the entity's property at `null` or `-1` as is currently the case when it isn't provided.

apiary.apib
dietrichm: Good question. I think this will have to be checked with UX and the Mobile team. Anyway the description is optional in the current client API.

@JorenSaey do you have wireframes and if so what is specified on the WorkOrder overview?

apiary.apib
dietrichm: I based this on feedback (https://github.com/teamleadercrm/api/pull/290#discussion_r255459660) from @stivni. Right now you have to _always_ provide address fields, so the address stays untouched on the WorkOrder entity (even when any of the customer's addresses you used changes over time).

You _can_ however make an optional link to a specific customer to make explicit that is an address of given customer.

apiary.apib
dietrichm: Product specifies this as the distance from the contractor's main address to the intervention address. They do not take into account separate distances per employee/assignee.

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+        + event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
+        + intervention_address (object, optional)
+            + Include Address
+            + customer (object, optional) - Customer this address refers to.
+                + type (enum[string], required)
+                    + Members
+                        + contact
+                        + company
+                + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + distance_travelled (object, optional)
```
jenssegers: Same here, this suggests that the work order is already done. Or is this something you fill in on completion? If so I don't think it should be part of draft.

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+            + type: `km` (enum[string], required)
```
jenssegers: PP: `unit`

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
+        + assignees (array, required)
+            + (object)
+                + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+                + hourly_rate (Money, optional)
+                + work_hours (array, optional)
+                    + (object)
+                        + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                        + duration: 3600 (number, required) - In seconds
+                        + description (string, optional)
+        + products (array, optional)
```
jenssegers: I guess some customers also have services as "products". But not sure if there's a good name that combines something like material/services.

Online templates often have different templates for companies that perform purely services, a combination of materials and services or simply a "description" of the work that is to be done or was done.

apiary.apib
JordyDeGroeve: WO are made on the job, no?
So this seems fine in that case.

apiary.apib
JorenSaey: I think we can do the same thing like TimeTracking: display it like (no description) in the ListView. If it's optional now, the next api should perhaps also make it optional. My opinion is that it is not for front end to decide whether something is required or not. We have other ways to show this ;)

apiary.apib
jenssegers: Looking at some online work order templates I find things like "expected start date", "starting date"

apiary.apib
jenssegers: Looking at some online work order templates I find things like "job location", "location", "client address"

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
...
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + customer (object, required)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + performed_at: `2017-05-26T10:01:49+00:00` (string, required)
```
jenssegers: Does this represent the date on which it should be performed or is this the date the person who performed it fills in when he did it? I think our work orders usually represent work to be done in the future? I think our properties should also reflect that.

Looking at some online work order templates I find things like "expected start date", "starting date", "scheduled start". Often in combination with the actual date on completion.

If it is the completion date it should not be part of the draft I think.

apiary.apib
```diff
@@ -3250,6 +3250,357 @@ Update the current timer. Only possible if there is a timer running.
 
 + Response 204
 
+# Group Work Orders
+
+## Work Orders [/workOrders]
+
+### workOrders.draft [POST /workOrders.draft]
```
jenssegers, line 3288: After reading a bit more about work orders I feel like this can also be `.schedule`. Eg. you schedule a work order, and complete it with details of the work you actually performed.

Possibly in combination with an `.assign` endpoint where you assign the scheduled work order to someone.

apiary.apib
dietrichm: To be checked. 👍

dietrichm:
Some property naming will still be checked with Product. However we also still have some room to discuss smaller things in the smaller API doc PRs to follow.

apiary.apib
jenssegers: From what I understand about work orders is that they are made up front and assigned to someone. And after the work is done, it is completed with more details. From that perspective, it's weird to have `distance_travelled` when the job is not started/completed yet.

jenssegers: Just checked with Ghislaine and a work order can be created up front, and you can optionally specify the distance you will travel to the location. And once you've done the job you can update that number with the amount you actually traveled.

apiary.apib
jenssegers: After short discussion, it seems that draft is ok.

src/01-general/work-orders.apib
```diff
@@ -61,8 +61,8 @@ Update a draft work order.
 + Request (application/json;charset=utf-8)
 
     + Attributes (object)
-        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string)
         + customer (object, required)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
```
dietrichm, line 50: Don't mind this line moving, it's from editing a hunk manually. Fixed in next commit.

apiary.apib
dietrichm: This now renamed to `location` and the customer link is removed.

apiary.apib
dietrichm: This is renamed to `distance`.

apiary.apib
dietrichm: This is renamed to `starts_at`.

apiary.apib
```diff
@@ -3281,6 +3281,327 @@ Update the current timer. Only possible if there is a timer running.
...
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + starts_at: `2017-05-26T10:01:49+00:00` (string, required)
+        + description: `Pipework intervention` (string, optional)
+        + department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional)
+        + layout_id: `a05320f6-b3ef-4ff9-a287-2074a12fb3e9` (string, optional)
+        + milestone_id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string, optional)
+        + event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional)
+        + location (Address, optional)
+        + distance (object, optional)
```
psrpinto: Q: Should we define the `distance` object in `datastructures.apib`?

apiary.apib
```diff
@@ -3281,6 +3281,327 @@ Update the current timer. Only possible if there is a timer running.
...
+        + work_hours (array, optional)
+            + (object)
+                + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + duration: 3600 (number, required) - In seconds
+                + description (string, optional)
+        + products (array, optional)
+            + (object)
+                + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
+                + quantity: `3` (number, optional)
```
WouterSioen: Q: do we want this to be optional? Isn't it more logical to require this instead of magically defaulting to 1 (I guess?)

apiary.apib
```diff
@@ -3281,6 +3281,327 @@ Update the current timer. Only possible if there is a timer running.
...
+        + distance (object, optional)
+            + unit: `km` (enum[string], required)
+                + Members
+                    + km
+            + value: `25.2` (number, required)
+        + work_hours (array, optional)
+            + (object)
+                + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+                + started_at: `2017-04-26T10:01:49+00:00` (string, required)
+                + duration: 3600 (number, required) - In seconds
```
WouterSioen, line 3313: Should we make this a VO with a unit? see https://github.com/teamleadercrm/api/pull/299

apiary.apib
dietrichm: Yup, should be required. Changing.

apiary.apib
dietrichm: Extracted it. There might be more possible data structures though.

psrpinto: ~Would it make sense to add `duration` to the `datastructures.apib`? I considered it but went with the safer option (not make it global) but since we're adding it here too, might make sense to make it a shared data structure.~

EDIT: Ignore, misread this thread

WouterSioen (dismissed):

tiagobastos (dismissed):

psrpinto (dismissed):
LGTM

apiary.apib
```diff
@@ -3281,6 +3281,311 @@ Update the current timer. Only possible if there is a timer running.
...
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+        + customer (object, optional)
+            + type (enum[string], required)
+                + Members
+                    + contact
+                    + company
+            + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+        + starts_at: `2017-05-26T10:01:49+00:00` (string, optional)
+        + description: `Pipework intervention` (string, optional, nullable)
+        + department_id: `cef01135-7e51-4f6f-a6eb-6e5e5a885ac7` (string, optional, nullable)
```
WouterSioen, line 3344: Q: do we want to be able to pass `null` as department id? I guess that if we draft a work order and we don't pass a department, the default department will be taken, and we don't want to be able to remove that one again?

apiary.apib
```diff
@@ -3281,6 +3281,311 @@ Update the current timer. Only possible if there is a timer running.
...
+        + layout_id: `a05320f6-b3ef-4ff9-a287-2074a12fb3e9` (string, optional, nullable)
```
WouterSioen, line 3345: Q: do we want to be able to pass `null` as layout id? I guess that if we draft a work order and we don't pass a layout, the default layout for the current department will be taken, and we don't want to be able to remove that one again?

apiary.apib
dietrichm: We actually allow the same for deals.update and do allow to "unset" the department again. Perhaps if a work order was drafted for a specific department and the client wants it to be linked to the default department instead (without knowing which one that is), it is possible by setting `department_id` to `null`.

apiary.apib
dietrichm: I'd use the same approach here as for the department ID.

apiary.apib
WouterSioen: Okay, that also seems to be the same way it is done in API v1

JordyDeGroeve (dismissed):

dietrichm:
Since we've been implementing parts and adding separate PRs that have been merged already, the conflicts are expected and will not be resolved. We will never merge this PR.

psrpinto (dismissed):
Conflict alert

JordyDeGroeve (dismissed):

WouterSioen (dismissed):

tiagobastos (dismissed):

jenssegers:
Going to remove myself as reviewer since this PR is not meant to be merged.

JorenSaey (dismissed):

JordyDeGroeve (dismissed):
Still not meant to be merged, but will approve the changes

tiagobastos (dismissed):

tiagobastos (dismissed):

JordyDeGroeve (dismissed):

mikeverf (dismissed):

apiary.apib
```diff
@@ -3487,10 +3487,11 @@ Get details for a single work order.
                         + type: `product` (string)
                         + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
                     + quantity: 3 (number)
-            + images (array)
+            + images (array) - Only included with request parameter `includes=images`.
                 + (object)
                     + id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string)
```
WouterSioen, line 3493: Q: should we nest this one deeper for variations of the same images e.g. thumbnails?

dietrichm: You mean `id` as a key in a new object, having an array of multiple objects with separate variations? That is possible but we mentioned this during the TA and @stivni made the remark we can perfectly return multiple objects with the same `id` but a different `variation`.

JordyDeGroeve (dismissed):

apiary.apib
dietrichm: Let's postpone this because we don't know if and when we will add variations (and sample base64 images).

apiary.apib
```diff
@@ -3281,6 +3281,312 @@ Update the current timer. Only possible if there is a timer running.
...
+                + (object)
+                    + product (object)
+                        + type: `product` (string)
+                        + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
+                    + quantity: 3 (number)
+            + images (array) - Only included with request parameter `includes=images`.
+                + (object)
+                    + id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string)
+                    + location: `https://cdn.teamleader.eu/file` (string) - A temporary URL where the image can be downloaded
+                    + expires: `2018-02-05T16:44:33+00:00` (string) - Expiration time of the temporary download URL
```
tiagobastos, line 3495: PP: Add a expires after the `created_at` of the WorkOrder.

JordyDeGroeve: What do you mean? A workOrder does not expire. It's the image link that expires.

tiagobastos: Not clear, sorry. 
The date on the `expires: 2018-02-05T16:44:33+00:00` is before the date the WorkOrder was created `created_at: 2019-05-09T11:25:11+00:00`. Should be after no? 
But, again, just a PP.

JordyDeGroeve: Aha! I will make sure this is fixed in the actual implementation.
This PR will never get merged.

apiary.apib
```diff
@@ -3281,6 +3281,312 @@ Update the current timer. Only possible if there is a timer running.
...
+
+### workOrders.deleteImage [POST /workOrders.deleteImage]
+
+Delete image from a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+        + image_id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string, required)
```
WouterSioen: Q: isn't one id enough? (the id of the work order image?)

dietrichm: If we keep both it's clear that you can't pass any file ID there and expect it to work. You need an actual WorkOrder ID.

Although it should be clear from the endpoint name as well. So not sure what to decide here.

WouterSioen: Both are okay for me, just seemed like obsolete data, but you're probably never going to have the file id without the work order id anyway

rathesDot (dismissed):

apiary.apib
```diff
@@ -3281,6 +3281,302 @@ Update the current timer. Only possible if there is a timer running.
...
+                + type (enum[string], required)
+                    + Members
+                        + contact
+                        + company
+                + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string, required)
+            + state: `draft`, `finalized` (enum[string], optional)
+                + (enum[string])
+                    + draft
+                    + finalized
+                    + sent
```
kevinwaelkens, line 3382: Q: Isn't this part incorrect? Renders like this:
Screenshot 2019-05-21 14 02 21

tiagobastos: You can check the https://github.com/teamleadercrm/api/compare/work-orders branch.
That one has the approved change for work orders.
This PR will never get merged by the way.

tiagobastos: But now that you mention this. 
The state is OK in the branch that I mentioned.

But the `sort.field` and `sort.order` have the same thing. 
But I can't seem to be able to use the correct syntax to fix it 😞

kevinwaelkens: That's really hard to find IMO... Isn't there a better way to look at this code? There's no PR for that branch open yet right now?

tiagobastos: Not yet. But you can checkout the branch and make a preview apiary.html file.

dietrichm: How Apiary renders is it admittedly weird. Not sure how we can fix this but it's an issue for multiple endpoints.

Since we revived an internal API endpoints repo we could add them there in the future and let the docs be rendered separately. If we have spare time we could/might move the ones in the integration branch already.

dietrichm:
To be moved to internal API.

PR #287, opened 2019-02-05
PRO-1380 Add billing method info to milestones.list

joaoinacio:
Implementation PR: https://github.com/teamleadercrm/core/pull/7956
E2E Tests: https://github.com/teamleadercrm/api-tests/pull/480

Updated to reflect naming decided for milestone invoice-ability info:
- changed from `billing_method` to `invoicing method` and include `non_invoiceable`:
https://github.com/teamleadercrm/core/pull/8264
- E2E Tests: https://github.com/teamleadercrm/api-tests/pull/514

JIRA: https://teamleader.atlassian.net/browse/PRO-1462

PR #291, opened 2019-02-12
PRO-1462: add invoiceable flag

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-1451

In the new project items list, we will need to visually mark uninvoiceable milestones. In order to do this, we will need to expose this information in the API.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2927,6 +2927,8 @@ Get a list of project milestones.
                     + Members
                         + open
                         + closed
+                + preferences (object)
+                    + invoiceable: true (boolean)
 
 ### milestones.info [GET /milestones.info]
 
@@ -2954,6 +2956,8 @@ Get details for a single milestone.
                 + Members
                     + open
                     + closed
+            + preferences (object)
+                + invoiceable: true (boolean)
 
 ### milestones.create [POST /milestones.create]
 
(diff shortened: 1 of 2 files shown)


jenssegers:
What does setting a milestone to non-invoiceable exactly do? Is that the same as setting the milestone to fixed-price of 0? @bbouwmeister

ketlane:
When invoicing a milestone, you can invoice according to time spent, or according to budget. If you choose "according to time spent", you can still create an invoice for fixed price milestones with 0 budget.

But if you set the "Invoiceable" flag to "No", you can't create an invoice for that milestone.

adognameddog (approved):

Kemosabert (approved):

WouterSioen (approved):

jenssegers (changes requested):
Sorry for blocking, but I think we're missing out on a different concept here. Briefly checked with @bbouwmeister and seems that we might need to give this some more thought?

adognameddog:
This can be closed since we agreed on another way of denoting a milestone's billing method.

ketlane:
See this PR in order to understand how to discover if a milestone is invoiceable or not: https://github.com/teamleadercrm/api/pull/287

PR #293, opened 2019-02-13
PRO-1464: add profit details to milestone

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-1425

image

In the new profit details design, the profitability details are shown at a milestone level so we need to share this information through the API. This structure mirrors the one at the project level.

### What has been done
- Added profit details at the milestone level

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2927,6 +2927,11 @@ Get a list of project milestones.
                     + Members
                         + open
                         + closed
+                + profit_info (object)
+                    + total_revenue (Money)
+                    + total_costs (Money)
+                    + profit (Money)
+
 
 ### milestones.info [GET /milestones.info]
 
@@ -2954,6 +2959,10 @@ Get details for a single milestone.
                 + Members
                     + open
                     + closed
+            + profit_info (object)
+                + total_revenue (Money)
+                + total_costs (Money)
+                + profit (Money)
 
 ### milestones.create [POST /milestones.create]
 
--- a/src/07-projects/milestones.apib
+++ b/src/07-projects/milestones.apib
@@ -52,6 +52,11 @@ Get a list of project milestones.
                     + Members
                         + open
                         + closed
+                + profit_info (object)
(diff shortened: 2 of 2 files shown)


ArnaudWeyts (approved):

WouterSioen (approved):

apiary.apib
```diff
@@ -2927,6 +2927,11 @@ Get a list of project milestones.
                     + Members
                         + open
                         + closed
+                + profit_info (object)
```
jenssegers, line 2930: Just wondering if there's a better name for this since it indicates both profit and loss. Is there a term that combines the two from a project/budget perspective?

jenssegers: Alternative: `profitability`

jenssegers: In a way, we're visualizing the current "budget progress".

ketlane: I like the `profitability` alternative. 

The problem is: if we change it here, I think we should also change it in the project.info call, before integrators start using this.

steftrenson: +1 for profitability https://en.wikipedia.org/wiki/Profit_margin

WouterSioen: Do we already return it in `project.info`? Otherwise I'd avoid still changing it without a BC layer. Once we pass it to customers, they can start depending on it, even when it's not documented.

adognameddog (commented):
Q: So... we changing it to `profitability`?

apiary.apib
ketlane: I'll just change it to `profitability` in milestones, then (at least for now). It seems to be the preferred term.

ketlane: I will also create a ticket to change `profit_info` to `profitability` in projects.info (with a BC layer). Not a priority atm though.

ketlane: https://teamleader.atlassian.net/browse/PRO-1468

PR #276, opened 2019-01-29
Add documentation for tasks

Jdww:
https://teamleader.atlassian.net/browse/CAL-1053

Adding documentation for the tasks api.

PR #299, opened 2019-02-18
Turn estimated_duration into an object with unit and value

psrpinto:
Following this discussion: https://github.com/teamleadercrm/api/pull/276#pullrequestreview-204258740

Once we agree on this, I will proceed to using it in the other documentation PRs we have open.

Note that the main `apiary.apib` file is not updated here because the Tasks section isn't being rendered, since we don't want to release it yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ lint: node_modules
 build: src node_modules
 	docker run -it --init --rm -v $(PWD):/code -w /code node:9 npm run build
 
-preview: node_modules
+preview: build
 	docker run -it --init --rm -v $(PWD):/code -w /code apiaryio/client preview --path=/code/apiary.apib --output=/code/apiary.html
 
 watch: node_modules
--- a/apiary.apib
+++ b/apiary.apib
@@ -314,7 +314,7 @@ This documentation only reflects the latest version of the API.
 If you need to make changes to your client or you want to make use of the latest feature, you might need to upgrade your API version.
 
 You can find your current API version, in the `X-Api-Version` header on any API response.
-The current latest version is **2018-10-30**.
+The current latest version is **2019-01-24**.
 
 After checking the API [changelog](#changelog) to see which endpoints work differently, you can upgrade your API version:
 
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -61,7 +61,11 @@ Create a new task.
         + description (string, required)
         + due_at: `2016-02-04T16:44:33+00:00` (string, required)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, required)
-        + estimated_duration: 3600 (number, optional) - In seconds
+        + estimated_duration (object, optional)
+            + unit: `s` (enum[string], required)
+                + Members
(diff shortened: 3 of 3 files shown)


rathesDot:
This looks a bit strange. Is the build missing something?

psrpinto:
@rathesDot I just added this to PR description: 

> Note that the main `apiary.apib` file is not updated here because the Tasks section isn't being rendered, since we don't want to release it yet.

WouterSioen (approved):

rathesDot (approved):

rathesDot:
Won't that lead to any problems when others try to build their things?

psrpinto:
@rathesDot Should have no impact, the `tasks.apib` file is just not being "imported" anywhere yet. Once we're ready to release, we'll import it in `src/apiary.apib`. 

We did this so that we could iterate on the API docs without actually releasing the Tasks endpoints. Once the docs are out there, the endpoints are officially released and we can no longer break backwards compatibility.

We plan on releasing these endpoints by the end of this sprint.

rathesDot:
Thanks for explaining @regularjack!

apiary.apib
```diff
@@ -314,7 +314,7 @@ This documentation only reflects the latest version of the API.
 If you need to make changes to your client or you want to make use of the latest feature, you might need to upgrade your API version.
 
 You can find your current API version, in the `X-Api-Version` header on any API response.
-The current latest version is **2018-10-30**.
+The current latest version is **2019-01-24**.
```
stivni, line 317: 👏 well spotted

src/08-tasks/tasks.apib
```diff
@@ -61,7 +61,11 @@ Create a new task.
         + description (string, required)
         + due_at: `2016-02-04T16:44:33+00:00` (string, required)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, required)
-        + estimated_duration: 3600 (number, optional) - In seconds
+        + estimated_duration (object, optional)
```
stivni, line 64: 👍 I like this a lot

stivni (commented):
~don't forget to build this~ scratch that

apiary.apib
WouterSioen: That's just the `build` that should've run on master 😬

stivni (approved):

Makefile
```diff
@@ -11,7 +11,7 @@ lint: node_modules
 build: src node_modules
 	docker run -it --init --rm -v $(PWD):/code -w /code node:9 npm run build
 
-preview: node_modules
+preview: build
```
tiagobastos, line 14: YES!! Then you can just do `make previews && open apiary.html` 😋

tiagobastos (approved):

dietrichm (approved):

regularjack:
DO NOT MERGE, waiting for implementation (I should have made it a draft PR).

r1dotl (approved):

psrpinto:
Merging since we need this for other PRs, even though the code hasn't yet been adapted to this format, but that's fine since this API isn't yet released.

PR #300, opened 2019-02-18
Current financial information of a project or milestone as "actuals"

adognameddog:
Actuals reflect the current financial state of a project or milestone. It distincts between revenue, costs and profit. This will not be released for now, but we need that data currently.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2656,7 +2656,7 @@ Projects allow users to work together as a team on a single project. They consis
 
 ### projects.list [GET /projects.list]
 
-Get a list of projects.
+Get a list of projects. "Actuals" is the current financial state of the project and is only accessible to account administrators and users with full admin access for projects.
 
 + Request (application/json;charset=utf-8)
 
@@ -2721,7 +2721,7 @@ Get a list of projects.
 
 ### projects.info [GET /projects.info]
 
-Get details for a single project.
+Get details for a single project. "Actuals" is the current financial state of the project and is only accessible to account administrators and users with full admin access for projects.
 
 + Request (application/json;charset=utf-8)
 
@@ -2892,7 +2892,7 @@ Every projects consists of one or more milestones which are limited in time and 
 
 ### milestones.list [GET /milestones.list]
 
-Get a list of project milestones.
+Get a list of project milestones. "Actuals" is the current financial state of the milestone and is only accessible to account administrators and users with full admin access for projects.
 
 + Request (application/json;charset=utf-8)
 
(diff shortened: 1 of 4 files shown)


ArnaudWeyts (dismissed):

nickwaelkens (dismissed):

joaoinacio (dismissed):

ketlane:
To me it's not very clear what we mean by "actuals" and "actuals.result" when reading the docs, but I understand we already agreed on this naming. Can we explain it in English words (maybe an example as well) in the documentation, though?

Maybe also add a note that they are only available to admins, so that integrators will not be puzzled when they can't access the information as regular users?

ketlane (dismissed):

apiary.apib
```diff
@@ -2713,11 +2713,15 @@ Get a list of projects.
                 + source (object)
                     + type: `deal` (string)
                     + id: `5023d7c2-80d7-4d4b-b2bd-0fcaa6a1f069` (string)
+                + actuals (object)
```
jenssegers: Should we add a comment next to this saying that it is only available for admins? I know it's in the description above, but would be nice to have it here as well 👍

jenssegers (dismissed):
Would just add a comment saying the property is only available for admins in the response documentation.

ketlane (dismissed):

jenssegers (dismissed):

joaoinacio (dismissed):

ArnaudWeyts (dismissed):

nickwaelkens (dismissed):

stivni:
@adognameddog I would recommend making an "integration"-branch for these changes, so we can merge this towards that branch...

src/07-projects/projects.apib
```diff
@@ -114,7 +118,10 @@ Get details for a single project.
                         + Members
                             + decision_maker
                             + member
-
+            + actuals (object) - Only available for account administrators and users with full admin access for projects.
```
stivni, line 121: Q: if only certain people can see this, will this be `null` in the other case or will the property be optionally returned? That choice should be documented...

stivni (commented):
also merge toward the (unexisting) integration branch?

Q : is this already in the live version? or only with a certain featureflag or include parameter?

stivni (commented):
@adognameddog if you can answer/fix my question, I think we should release this

ArnaudWeyts (approved):

adognameddog:
I think an integration branch is no longer necessary 🙈

mikeverf (approved):

adognameddog:
The documentation of this actually (pun very much intended) changed. I'll create a new PR.

PR #301, opened 2019-02-18
Add estimated_duration and work_type to tasks.info and tasks.list.

r1dotl:
Implementation: https://github.com/teamleadercrm/core/pull/8261

## Note
Since we're not ready to release these endpoints yet, we decided to not include the `tasks.apib` in the `src/apiary.apib` file. That way we can iterate on the docs through multiple PRs. Once we're ready to release, we will include `tasks.apib` in `src/apiary.apib`.

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -31,6 +31,10 @@ Get a list of tasks.
                 + completed: false (boolean)
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
+                + estimated_duration: 3600 (number) - In seconds
+                + work_type (object, nullable)
+                    + type: `workType` (string)
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
 
 ### tasks.info [GET /tasks.info]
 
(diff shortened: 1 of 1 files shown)


src/08-tasks/tasks.apib
```diff
@@ -31,6 +31,10 @@ Get a list of tasks.
                 + completed: false (boolean)
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
+                + estimated_duration: 3600 (number) - In seconds
```
psrpinto: B: We should use an object for duration fields from now on. See https://github.com/teamleadercrm/api/pull/299

r1dotl: I know, but i though that change would be made after the code was merged. Since the current code of tasks.list and tasks.info doesn't use the object yet.

psrpinto: The code doesn't use the object yet indeed, but the documentation is using the new format everywhere except here, unless I'm mistaken. At the moment the code does not conform with the documentation, but I'm taking care of that.

src/08-tasks/tasks.apib
```diff
@@ -31,6 +31,14 @@ Get a list of tasks.
                 + completed: false (boolean)
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
+                + estimated_duration (object, optional)
+                    + unit: `s` (enum[string], required)
+                        + Members
+                            + s - Seconds
+                    + value: `60` (number, required)
+                + work_type (object, nullable)
```
WouterSioen, line 39: Why is this one `nullable` and the other one `optional`? I guess it should both be `nullable`? Since we always pass the same fields, no matter what in read endpoints?

r1dotl: Because i copied it from a Request and this is the Reponse :p
Just removed the optional and required since it makes no sense in a response.

WouterSioen (approved):

psrpinto (approved):

joaoinacio (approved):

PR #303, opened 2019-02-19
Document the filter on term option in deals.list

Kemosabert:
### ADDED
Option to filter on term in the `deals.list` endpoint

Note: I found the keys on which it actually searches in the `DbalDealRepository.php` file.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1349,6 +1349,7 @@ Get a list of deals.
     + Attributes (object)
         + filter (object, optional)
             + ids: `3f507137-e599-41d4-8625-0e2adba0be4c`,`ae97f6d3-0bb4-43d3-b13d-f06069d07e98` (array[string], optional)
+            + term: `Interesting deal` (string, optional) - Filters on the title, reference and customer's name
             + customer (object, optional)
                 + type (enum[string], required)
                     + Members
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -1349,6 +1349,7 @@ Get a list of deals.
     + Attributes (object)
         + filter (object, optional)
             + ids: `3f507137-e599-41d4-8625-0e2adba0be4c`,`ae97f6d3-0bb4-43d3-b13d-f06069d07e98` (array[string], optional)
+            + term: `Interesting deal` (string, optional) - Filters on the title, reference and customer's name
```
WouterSioen, line 1352: I think `quotation number` is a more correct domain term than `reference`, right?

Summoning the deals team: @dietrichm @steftrenson

Kemosabert: Ah, I went with `reference` as I assumed that's the key it's refering to

WouterSioen: Going to double check if that's the same.

WouterSioen: Yep, it's the same!

WouterSioen (approved):

mrtus (approved):

dietrichm (approved):

dietrichm:
✅ Checked

JordyDeGroeve (approved):

PR #307, opened 2019-02-20
Removed responsible_user_id from apib.

r1dotl:
For now we won't support responsible_user_id on the public API.
We will have to figure out how we're going to proceed regarding the teams on tasks, after that we will put it public.


WouterSioen (approved):

psrpinto:
Closing since we're making some changes concerning `responsible` of a task.

PR #309, opened 2019-02-21
estimated_duration of Tasks must be in minutes

psrpinto:
Even though we want to standardise on seconds everywhere, it just doesn't make sense here since in the Web UI we're using minutes. Using seconds on the API would lead to unexpected behaviour since we'd have to round to minutes.

Even though there is no [International System of Units](https://en.wikipedia.org/wiki/International_System_of_Units) unit for minutes, the ad-hoc "standard" is `min`, apparently. There's more information on that wikipedia page.

https://github.com/teamleadercrm/core/pull/8350

## Note
Since we're not ready to release these endpoints yet, we decided to not include the `tasks.apib` in the `src/apiary.apib` file. That way we can iterate on the docs through multiple PRs. Once we're ready to release, we will include `tasks.apib` in `src/apiary.apib`.

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -35,9 +35,9 @@ Get a list of tasks.
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
                 + estimated_duration (object)
-                    + unit: `s` (enum[string])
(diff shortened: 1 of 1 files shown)


r1dotl (approved):
:ok_hand:

WouterSioen (approved):

dcsg (approved):

PR #310, opened 2019-02-21
Adapt API docs for new assignee format

r1dotl:
Teams are not part of this PR, but we will add it later.
 
- tasks.list
	- `user_id` filter - be able to get tasks where the `user_id` is the responsible or is part of the responsible team.
	- response returns the `assignee` object.
- tasks.info
	- response returns the `assignee` object.
- tasks.create
	- accept `assignee` object instead of `assignee_id`
	- only accept assignee of type `user`
- tasks.update
	- accept `assignee` object instead of `assignee_id`
	- only accept `assignee` of type `user`

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -101,6 +101,7 @@ Create a new task.
             + type: `user` (enum[string])
                 + Members
                     + user
+                    + team
             + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
 
 + Response 201 (application/json;charset=utf-8)
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -8,7 +8,7 @@ Get a list of tasks.
 
     + Attributes (object)
         + filter (object, optional)
-            + assignee_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional)
+            + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional)
             + completed: true (boolean, optional)
         + page (Page, optional)
         + sort (array, optional)
@@ -42,6 +42,12 @@ Get a list of tasks.
                 + work_type (object, nullable)
                     + type: `workType` (string)
                     + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + responsible (object)
+                    + type: `user` (enum[string])
+                        + Members
+                            + user
+                            + team
+                    + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
 
 ### tasks.info [GET /tasks.info]
 
@@ -69,6 +75,12 @@ Get information about a task.
             + work_type (object, nullable)
                 + type: `workType` (string)
                 + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+            + responsible (object)
+                + type: `user` (enum[string])
+                    + Members
+                        + user
+                        + team
+                + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
 
 ### tasks.create [POST /tasks.create]
 
@@ -85,7 +97,11 @@ Create a new task.
                 + Members
                     + s - Seconds
             + value: `60` (number, required)
-        + assignee_id: `445adf2b-9b76-077b-8b52-9c1c7c32a601` (string, optional)
+        + responsible (object, optional)
+            + type: `user` (enum[string])
+                + Members
+                    + user
+            + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
 
 + Response 201 (application/json;charset=utf-8)
 
@@ -110,7 +126,11 @@ Update a tasks.
                 + Members
                     + s - Seconds
             + value: `60` (number, required)
-        + assignee_id: `445adf2b-9b76-077b-8b52-9c1c7c32a601` (string, optional)
+        + responsible (object, optional)
+            + type: `user` (enum[string])
+                + Members
+                    + user
+            + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
 
 + Response 204
 
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -130,6 +130,7 @@ Update a tasks.
             + type: `user` (enum[string])
                 + Members
                     + user
+                    + team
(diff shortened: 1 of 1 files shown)


psrpinto (dismissed):

src/08-tasks/tasks.apib
```diff
@@ -42,6 +42,12 @@ Get a list of tasks.
                 + work_type (object, nullable)
                     + type: `workType` (string)
                     + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + responsible (object)
```
stivni: tbh, now that I see this, why don't call it "assigned to" like in the UI?

r1dotl: > tbh, now that I see this, why don't call it "assigned to" like in the UI?

Before this we had assignee id, but we checked and in the rest of the app the "term" assignee was not used anywhere, they used "responsible user", so to keep everything with the same naming, since the meaning is the same (the only difference was the naming), we decided to change to responsible_user. After that we decided that the responsible_user would not work, because it could be a user or a team so we created a VO for Responsible with a type and id, so we could easily manage the user and team type. If we change this to "assigned to" like the UI, we will introduce a new term on the API and it will be used only, because the UI has it as "assigned to" while the rest of the app uses the term "responsible user" for this type of situation.

psrpinto (dismissed):

psrpinto (dismissed):

WouterSioen (dismissed):

src/08-tasks/tasks.apib
jenssegers: If "assigned to" makes more sense we can also change the UI. We're doing that in other places as well. For example; in payments, we're introducing "seats" on your license that will also be used in the UI.

jenssegers: My UI shows assigned to by the way:

image

psrpinto: I'm convinced we should change this to `assignedTo`. `assignee` was weird, but `assignedTo` makes sense to me.

psrpinto (dismissed):

mrtus (dismissed):

WouterSioen (dismissed):

src/08-tasks/tasks.apib
```diff
@@ -42,6 +42,12 @@ Get a list of tasks.
                 + work_type (object, nullable)
                     + type: `workType` (string)
                     + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + assigned_to (object)
```
jenssegers: Comparing to other API endpoints it would seem that `assignee` would be a better fit. See `invoicee`, `creator`, `customer`, ...

Does that make sense here too?

src/08-tasks/tasks.apib
```diff
@@ -8,7 +8,7 @@ Get a list of tasks.
 
     + Attributes (object)
         + filter (object, optional)
-            + assignee_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional)
+            + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional)
```
jenssegers: Let's make this filter similar to the one on deals/projects https://developer.teamleader.eu/#/reference/deals/deals/deals.list

So an `assignee` filter with a type and id.

jenssegers (dismissed):
Change `user_id` filter to `assignee` with a type/id.

Would also rename `assigned_to` to `assignee` unless that does not make sense in the domain.

src/08-tasks/tasks.apib
psrpinto: This isn't filtering by "assignee" though, it also returns tasks that are assigned to a team to which you belong to. This is consistent with how `events.list` works, with a `user_id` filter that returns events for which you are attending.

regularjack:
@jenssegers  Aren't filters supposed to be `foo_id` instead of a `foo` object? I'm confused.

src/08-tasks/tasks.apib
jenssegers: If we want to add a filter to search on assignee with type/id, how will that relate to this `user_id`, and will that not be confusing for the people using it?

psrpinto: > If we want to add a filter to search on assignee with type/id, how will that relate to this `user_id`, and will that not be confusing for the people using it?

The idea was that that would be a different filter, but not sure that makes sense.

What we need is a filter that returns both tasks that are assigned to you AND tasks that are assigned to a team to which you belong to. You should be able to get all those tasks in one request.

I think both this and the "filter to search on assignee with type/id" are valid use cases, but for the moment we only need the first one. 

Not sure what to do.

r1dotl:
I think it makes sense like it is right now, since the idea here is to return both tasks where the user is the responsible and the user is part of the responsible team.

src/08-tasks/tasks.apib
jenssegers: If you think it does not make sense to use the assignee filter we can use the `user_id` with a proper comment next to it. We can still make a breaking change later I guess.

psrpinto:
@jenssegers We'll keep the filter as `user_id` for now. If in the future we decide to have an `assignee` filter, we can keep supporting the `user_id` one but remove it from the docs.

About renaming `assigned_to` to `assignee`, we can do it. But are you absolutely sure about it? We went from `assignee` to `responsible` to `assigned_to` and now back to `assignee` so, before we change the code for the 4th time, we just want to be sure :)

src/08-tasks/tasks.apib
psrpinto: We'll add a comment to make clear what the `user_id` filter does.

psrpinto:
Should be good to go now

src/08-tasks/tasks.apib
```diff
@@ -8,7 +8,7 @@ Get a list of tasks.
 
     + Attributes (object)
         + filter (object, optional)
-            + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional)
+            + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
```
r1dotl: Q: should this be `...this user or to a team...` or `...this user and to a team...`?

geoffreydhuyvetters (approved):

jenssegers (approved):

psrpinto (approved):

PR #313, opened 2019-02-25
Add tax rate uuid to quotation.info endpoint

Yarn-e:
Bookkeeping integrations need to have more details regarding the Tax Rate. The percentage is not enough since the same percentage value can have different meanings. By exposing the Tax Rate ID we give them the possibility to fetch more details of that given Tax Rate.

This change implies the removal of the `rate` from the `tax` object in `quotations.info` response. In exchange, we are adding the `id` and the `type` to that object. That way Integrators will be able to fetch the details of the Tax Rate.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -314,7 +314,7 @@ This documentation only reflects the latest version of the API.
 If you need to make changes to your client or you want to make use of the latest feature, you might need to upgrade your API version.
 
 You can find your current API version, in the `X-Api-Version` header on any API response.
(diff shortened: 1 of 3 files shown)


EOSullivanBerlin (approved):

jdrieghe (changes requested):
We should still
- document the BC break
- build the `apiary` file

steftrenson (approved):

luisaterrabrito (approved):

PR #318, opened 2019-02-27
Allow passing time when drafting a work order

WouterSioen:


tiagobastos (approved):

WouterSioen:
Added `HOLD` since we need to decide if we're going to allow work hours on multiple days or not.

dietrichm (approved):

tiagobastos (approved):

apiary.apib
```diff
@@ -3374,6 +3374,12 @@ Add a draft work order.
                + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
                + quantity: 3 (number, required)
        + location (Address, optional)
+       + work_hours (array, optional)
+           + (object)
+               + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+               + started_at: `2019-02-27T10:01:49+00:00` (string, required)
+               + duration (Duration, required)
```
mrtus: Q: Should we have an end time instead? It's easier to say when you started & ended than to calculate the amount of time you worked.

dietrichm: We checked this with the mobile team and they prefer this format (I guess this is what they've been using already for time tracking). Since we need to move fast with WorkOrders we're just accepting this format for now.

JordyDeGroeve (approved):
Any update on this?

apiary.apib
jenssegers: If we want to offer the same functionality as time tracking can we actually make it the same? Because now we're introducing a duration object while time tracking just has a duration in seconds.

My PP would be to offer options with duration and start/end like `timeTracking.add`.

jenssegers (changes requested):
Would like to make adding work hours similar to adding time tracking.

(Missed this PR because of the hold label by the way 🙈)

dietrichm:
To be discussed in meeting.

apiary.apib
JordyDeGroeve: We will keep it in sync with the time tracking options.
So you can provide duration as an integer. 
Later on, we can add the start/end time option.

dietrichm:
Removing a duplicate base branch closed this one. It shouldn't have had a base other than `master` anyway.

JordyDeGroeve (dismissed):

dietrichm (dismissed):

tiagobastos (dismissed):

dietrichm (approved):

apiary.apib
```diff
@@ -3374,6 +3374,12 @@ Add a draft work order.
                + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
                + quantity: 3 (number, required)
        + location (Address, optional)
+       + work_hours (array, optional)
+           + (object)
+               + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+               + started_at: `2019-02-27T10:01:49+00:00` (string, required)
+               + duration: 3600 (number, required) - In seconds
```
tiagobastos, line 3544: Will we change this to match TimeTracking format? 
`started_at` & `duration` OR  `started_at` & `ended_at`

tiagobastos (approved):

apiary.apib
dietrichm: No. This was in the initial proposal but we choose to not implement this immediately. The mobile team only uses the duration.

jenssegers (dismissed):

src/09-workorders/work-orders.apib
```diff
@@ -23,6 +23,12 @@ Add a draft work order.
                + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
                + quantity: 3 (number, required)
        + location (Address, optional)
+       + work_hours (array, optional)
```
jenssegers: Should we rename this to `time` so that we have time & materials? Would be nice 😬

tiagobastos: 🤯

tiagobastos (approved):

JordyDeGroeve (approved):

apiary.apib
JorenSaey: For TimeTracking it is indeed just duration in seconds, I don't know why this is an Object here.
Basically I agree with Jens to make it the same as TimeTracking but for simplicity perhaps only do the started_at + duration combination instead of everything (started_at + ended_at, ended_at + duration, ...)

jenssegers (approved):

apiary.apib
JordyDeGroeve: Indeed. The object has been removed. And as discussed with mobile, we only implemented the first combination due to deadlines.

PR #319, opened 2019-02-28
Added a suggestion for the purchase price to the documentation

EOSullivanBerlin:


luisaterrabrito (dismissed):

src/06-products/products.apib
```diff
@@ -24,3 +24,5 @@ Get a list of products.
                         + name: `cookies` (string)
                         + description: `dark chocolate` (string)
                 + external_id: `123` (string)
+                + purchase_price (Money)
```
dcsg: I believe you forgot to build the documentation it should also add it to the `https://github.com/teamleadercrm/api/blob/master/apiary.apib`

dcsg (changes requested):

src/06-products/products.apib
EOSullivanBerlin: Its built now.

dcsg (dismissed):

luisaterrabrito (dismissed):

Yarn-e (dismissed):

jdrieghe (dismissed):

dcsg (dismissed):

src/06-products/products.apib
```diff
@@ -24,3 +24,5 @@ Get a list of products.
                         + name: `cookies` (string)
                         + description: `dark chocolate` (string)
                 + external_id: `123` (string)
+                + purchase_price (Money, optional)
```
dcsg, line 27: I believe this can be `null` but not `optional`

EOSullivanBerlin: Okay what should we do? I am a bit lost.

PR #324, opened 2019-03-04
Publish Tasks API docs

psrpinto:
Just adding the `tasks.apib` file to `src/apiary.apib` so that the Tasks docs are released. All this has already been reviewed.


r1dotl (approved):
:ok_hand:

tiagobastos (approved):

dietrichm (approved):

PR #329, opened 2019-03-08
Add workOrders.info endpoint

dietrichm:
Adding documentation for the WorkOrders .info endpoint with the currently available properties.


apiary.apib
```diff
@@ -3518,6 +3518,52 @@ Update the current timer. Only possible if there is a timer running.
 
 ## Work Orders [/workOrders]
 
+### workOrders.info [GET /workOrders.info]
+
+Get details for a single work order
```
JordyDeGroeve: PP: End a sentence with a dot.

apiary.apib
```diff
@@ -3518,6 +3518,52 @@ Update the current timer. Only possible if there is a timer running.
...
+                + id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string)
+            + location (Address)
+            + created_at: `2019-05-09T11:25:11+00:00` (string)
+            + updated_at: `2019-05-09T11:30:58+00:00` (string)
+            + materials (array)
+                + (object)
+                    + product (object)
+                        + type: `product` (string)
+                        + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
+                    + quantity: `3` (number)
```
JordyDeGroeve: PP: numbers don't need to be encapsulated by quotes/backticks

jenssegers (dismissed):

apiary.apib
JordyDeGroeve: 😉

JordyDeGroeve (dismissed):

src/09-workorders/work-orders.apib
```diff
@@ -39,6 +39,14 @@ Get details for a single work order.
...
             + created_at: `2019-05-09T11:25:11+00:00` (string)
             + updated_at: `2019-05-09T11:30:58+00:00` (string)
+            + work_hours (array)
+                + (object)
+                    + user (object)
+                        + type: `user` (string)
+                        + id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string)
+                    + started_at: `2019-02-27T10:01:49+00:00` (string)
+                    + duration: 3600 (number) - In seconds
+                    + description (string)
```
JordyDeGroeve: Q: no example for the description?

dietrichm: The draft endpoint doesn't provide one either. So let's keep it the same since it ought to be showing info for the same one.

tiagobastos (dismissed):

src/09-workorders/work-orders.apib
JordyDeGroeve: Would update both of them. These examples/placeholders are helpful for our integrators.

apiary.apib
```diff
@@ -3518,6 +3518,60 @@ Update the current timer. Only possible if there is a timer running.
...
+                    + draft
+                    + finalized
+                    + sent
+            + customer (object)
+                + type (enum[string])
+                    + Members
+                        + contact
+                        + company
+                + id: `09e5d75f-f817-4872-9723-57fbb8ff710d` (string)
+            + starts_at: `2017-05-26T10:01:49+00:00` (string)
```
mrtus, line 3546: Q: This feels a bit weird, what if the work order was already completed?

JordyDeGroeve (approved):

apiary.apib
```diff
@@ -3564,7 +3564,7 @@ Get details for a single work order.
                         + id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string)
                     + started_at: `2019-02-27T10:01:49+00:00` (string)
                     + duration: 3600 (number) - In seconds
```
tiagobastos, line 3566: Will we change this to match TimeTracking format? 
`started_at` & `duration` OR  `started_at` & `ended_at`

tiagobastos (approved):

apiary.apib
JordyDeGroeve: Should be in the same tense as the `started_at` of the work_hours imo

apiary.apib
dietrichm: No. This was in the initial proposal but we choose to not implement this immediately. The mobile team only uses the duration.

apiary.apib
tiagobastos: We already name this parameter like this on `workOrders.draft`. 
If we change it here, to be consistent, we should name it there also. 
But altering it now would be a BC.

dietrichm: We discussed this during (long) grooming sessions and this is the most non-restrictive term. Using a past tense doesn't make sense when you draft work orders to be executed days later. Work hours _may_ be drafted as well, but there we took the format from the time tracking endpoints.

apiary.apib
```diff
@@ -3518,6 +3518,60 @@ Update the current timer. Only possible if there is a timer running.
...
+            + milestone (object)
+                + type: `milestone` (string)
+                + id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string)
+            + event (object)
+                + type: `event` (string)
+                + id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string)
+            + location (Address)
+            + created_at: `2019-05-09T11:25:11+00:00` (string)
+            + updated_at: `2019-05-09T11:30:58+00:00` (string)
+            + work_hours (array)
```
jenssegers: Same remark as in draft, should we rename this to `time`?

tiagobastos (approved):

JordyDeGroeve (approved):

jenssegers (approved):

PR #330, opened 2019-03-08
[CORE-1778] Update draft workorder

tiagobastos:
Adding documentation for the `workOrders.update` endpoint with the currently available properties.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3549,6 +3549,29 @@ Add a draft work order.
             + type: `workOrder` (string)
             + id: `7c1d8672-f502-4333-9ea4-7a45add15115` (string)
 
+### workOrders.update [POST /workOrders.update]
+
+Update a draft work order.
+
++ Request (application/json;charset=utf-8)
+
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -3549,6 +3549,29 @@ Add a draft work order.
             + type: `workOrder` (string)
             + id: `7c1d8672-f502-4333-9ea4-7a45add15115` (string)
 
+### workOrders.update [POST /workOrders.update]
+
+Update a draft work order.
```
JordyDeGroeve, line 3554: PP: `Update a drafted work order.`

dietrichm (approved):

jenssegers (approved):

apiary.apib
dietrichm: I think it's fine like this though. "Draft invoice" or "Draft work order" is a valid way to describe it.

JordyDeGroeve (approved):

PR #333, opened 2019-03-13
Add guidelines to deal with user permissions for specific properties in objects

dcsg:
This proposal wants to tackle the need to handle User Permissions for specific properties of an Object when creating, updating, or fetching it.

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -171,7 +171,7 @@ Example:
 }
 ```
 
-### User Permissions
+### Restricted properties
 
 #### Fetching objects
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -173,7 +173,7 @@ Example:
 
 ### Restricted properties
 
-When a User don't have permissions to access a specific property the endpoint **MUST**:
+When a User doesn't have permissions to access a specific property the endpoint **MUST**:
 
  - Not return that **property** in the response.
  - When updating or creating an Object and it contains that **property**, then the endpoint **MUST** return a `403` error response with an appropriate explanation. The property **MUST** be optional.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -176,7 +176,7 @@ Example:
 When a User doesn't have permissions to access a specific property the endpoint **MUST**:
 
  - Not return that **property** in the response.
- - When updating or creating an Object and it contains that **property**, then the endpoint **MUST** return a `403` error response with an appropriate explanation. The property **MUST** be optional.
+ - When writing to a resource that contains that **property**, the endpoint **MUST** return a `403` error response with an appropriate message.
  - The property **MUST** be documented in API docs as ***optional*** for both cases.
 
 ### Date and time
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -72,7 +72,7 @@ We try to interpret HTTP Status codes the right way:
  - `204` - No Content, on resource updates or actions
  - `400` - Bad Request, the request contains invalid data or references non-existing resources
  - `401` - Unauthorized, invalid or missing access token
- - `403` - Forbidden, not allowed to access this resource
+ - `403` - Forbidden, not allowed to access this resource, or a specific property of a resource
  - `404` - Not Found, resource not found
  - `429` - Too Many Requests, client has reached the API rate limit
  - `500` - Internal Server Error, something went wrong on our end
@@ -170,6 +170,18 @@ Example:
   }
 }
 ```
+
+### User Permissions
+
+#### Fetching objects
+
+When fetching objects if the User has permissions to fetch it but he lacks a specific property permission then that property will not be returned in the response.
+
+#### Creating/Updating objects
+
+A User that has permission to add a Object might not have permissions to certain properties of that object. In the case 
+he tries to add or update a object with a property that he does not have access to, the request will fail and will 
+return a `403` response with a message explaining what caused it.
 
 ### Date and time
 
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -72,7 +72,7 @@ We try to interpret HTTP Status codes the right way:
  - `204` - No Content, on resource updates or actions
  - `400` - Bad Request, the request contains invalid data or references non-existing resources
  - `401` - Unauthorized, invalid or missing access token
- - `403` - Forbidden, not allowed to access this resource, or a specific property of a resource
+ - `403` - Forbidden, not allowed to access this resource or update a property you are not allowed to change
  - `404` - Not Found, resource not found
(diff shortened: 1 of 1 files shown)


GUIDELINES.md
```diff
@@ -72,7 +72,7 @@ We try to interpret HTTP Status codes the right way:
  - `204` - No Content, on resource updates or actions
  - `400` - Bad Request, the request contains invalid data or references non-existing resources
  - `401` - Unauthorized, invalid or missing access token
- - `403` - Forbidden, not allowed to access this resource
+ - `403` - Forbidden, not allowed to access this resource, or a specific property of a resource
```
dcsg: We thought about returning a `400` when the User don't have permissions to access a specific property but it would be incorrect because the request is not malformed nor invalid, is indeed a matter of authorization.

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
 }
 ```
 
+### User Permissions
+
+#### Fetching objects
+
+When fetching objects if the User has permissions to fetch it but he lacks a specific property permission then that property will not be returned in the response.
```
dcsg: We discussed the possibility to simple return `null` instead of making the properties `optional` in the response. If we would make them `null` that would mean that the integrator wouldn't know if the property doesn't have value or if the User don't have the right to access it. Making it optional will make it explicit that the User don't have access to it but will make it less friendly for our integrators because they will have to have some if conditions for those properties.

steftrenson (dismissed):

adognameddog (dismissed):

jdrieghe (dismissed):

jdrieghe:
Good enough to 🚢 !

jdrieghe:
Thanks for the efforts, @dcsg 🙌

GUIDELINES.md
jenssegers: When reading this, this would also return a 403 when retrieving info about a resource with a property you don't have access to. Would change it to:

```suggestion
 - `403` - Forbidden, not allowed to access this resource or update a property you are not allowed to change
```

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
 }
 ```
 
+### User Permissions
```
jenssegers: This could be under a section called:

```suggestion
### Restricted properties
```

GUIDELINES.md
jenssegers: Would also not split up the reading and writing since they're both related to restricted properties.

```suggestion
Sometimes users don't have the required permissions to read or write certain data. In our API responses, we should not return the properties the user has no access to.

If that user tries to change a property he is not allowed to change, then we will return a `403` error response with an appropriate explanation.
```

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
 }
 ```
 
+### Restricted properties
+
+#### Fetching objects
+
+When fetching objects if the User has permissions to fetch it but he lacks a specific property permission then that property will not be returned in the response.
```
psrpinto:
```suggestion
When fetching objects, if the User has permissions to fetch but lacks a specific property permission, that property will not be returned in the response.
```

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
...
+
+#### Creating/Updating objects
+
+A User that has permission to add a Object might not have permissions to certain properties of that object. In the case
```
psrpinto:
```suggestion
A User that has permission to add an Object might not have permissions to certain properties of that object. In case 
```

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
...
+### Restricted properties
+
+#### Fetching objects
+
+When fetching objects if the User has permissions to fetch it but he lacks a specific property permission then that property will not be returned in the response.
+
+#### Creating/Updating objects
+
+A User that has permission to add a Object might not have permissions to certain properties of that object. In the case 
+he tries to add or update a object with a property that he does not have access to, the request will fail and will
```
psrpinto:
```suggestion
they try to add or update an Object with a property that they don't have access to, the request will fail and 
```

GUIDELINES.md
```diff
@@ -171,6 +171,18 @@ Example:
...
+
+#### Fetching objects
+
+When fetching objects if the User has permissions to fetch it but he lacks a specific property permission then that property will not be returned in the response.
+
+#### Creating/Updating objects
+
+A User that has permission to add a Object might not have permissions to certain properties of that object. In the case 
+he tries to add or update a object with a property that he does not have access to, the request will fail and will 
+return a `403` response with a message explaining what caused it.
```
psrpinto:
```suggestion
a `403` response with an explanatory message will be returned.
```

psrpinto (commented):
Suggesting some phrasing improvements. Also, I agree with most other remarks.

stivni (commented):
General remark:

It feels to me that we are trying say what "will" be the case, as if the intended audience of this page are the API users...
But I think the guidelines should be aimed at us, the creators of the API...
So it could even be in some kind of RFC style: You MUST ... You SHOULD ... You CAN ...

dcsg:
@jenssegers @regularjack @stivni I marked everything as resolved because I changed the approach to the text to be in the RFC style. Could you review it again?

GUIDELINES.md
```diff
@@ -171,6 +171,14 @@ Example:
 }
 ```
 
+### Restricted properties
+
+When a User don't have permissions to access a specific property the endpoint **MUST**:
```
psrpinto:
```suggestion
When a User doesn't have permissions to access a specific property the endpoint **MUST**:
```

psrpinto (dismissed):
👍

psrpinto (dismissed):

GUIDELINES.md
```diff
@@ -171,6 +171,14 @@ Example:
 }
 ```
 
+### Restricted properties
+
+When a User doesn't have permissions to access a specific property the endpoint **MUST**:
+
+ - Not return that **property** in the response.
+ - When updating or creating an Object and it contains that **property**, then the endpoint **MUST** return a `403` error response with an appropriate explanation. The property **MUST** be optional.
```
jdrieghe:
```suggestion
 - When writing to a resource that contains that **property**, the endpoint **MUST** return a `403` error response with an appropriate message.
```

stivni (dismissed):

adognameddog (dismissed):

steftrenson (approved):

dietrichm (approved):

adognameddog (approved):

psrpinto (approved):

psrpinto:
🚢

PR #335, opened 2019-03-20
Purchase order number for invoicing end points 

EOSullivanBerlin:
Added the documentation for the planned PO number end points.


Yarn-e (dismissed):

lowiebenoot (dismissed):
Should we add it to invoices.list as well?

EOSullivanBerlin:
@lowiebenoot Pieter wants us to be able to filter on the PO number with the invoice.list. I can either add it now or make another PR later?

dcsg (dismissed):

jenssegers (dismissed):

lowiebenoot:
Adding it to the list already makes sense to me, filtering can be added later.

EOSullivanBerlin:
Okay I will add it.

jenssegers (approved):

steftrenson (commented):
Since the value can also be a string, should we go with purchase_order_number or just purchase_order?

EOSullivanBerlin:
I think number is still fine. For strings made up of Letters characters and numbers. 
1234-24343-AB I would still  call that a number for example.

lowiebenoot (approved):
Numbers often contain alphabetic characters, so I would just leave it like that (e.g. vat number, house number, ...)

luisaterrabrito (approved):

Yarn-e (approved):

PR #341, opened 2019-04-03
Add proposal for invoices.send

luisaterrabrito:


apiary.apib
```diff
@@ -2425,6 +2425,20 @@ Register a payment for an invoice.
...
 
+### invoices.send [POST /invoices.send]
+
+Send an invoice.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `7abb325c-e063-42a4-8fb4-1b730759645a` (string, required)
+        + to: `mary@teamleader.eu`, `john@teamleader.eu` (array[string], required)
```
dcsg, line 2436: Q: Should we also include the `cc` and `bcc` in this proposal? In the UI is possible to set those as well.

apiary.apib
```diff
@@ -2425,6 +2425,20 @@ Register a payment for an invoice.
...
+        + subject: `compliments` (string, required)
+        + content: `amazing compliments` (string, required)
```
dcsg, line 2438: Q: would it also make sense to accept the template to be used for the email?

luisaterrabrito: that is for later

luisaterrabrito:
Still to be discussed

PR #344, opened 2019-04-09
[PRO-1726] API documentation for adding a product reference to project items

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-1726

Implementation PR: https://github.com/teamleadercrm/core/pull/9255
E2E tests PR: https://github.com/teamleadercrm/api-tests/pull/591

In the case of project items that don't have a purchase price set, we want to suggest the purchase price of the referenced product (where applicable). So we're going to expose the related product, so that they could be sideloaded.

### What has been done
- Added a product reference to project items

### DIFF
--- a/src/07-projects/project-items.apib
+++ b/src/07-projects/project-items.apib
@@ -70,6 +70,9 @@ Get a list of project items.
                 + total
                     + internal_cost (Money)
                     + billable_price (Money)
+                + product (object, nullable)
+                    + type: `product` (string)
+                    + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
                 + added_at: `2018-12-04T16:44:33+00:00` (string)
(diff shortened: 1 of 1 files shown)


adognameddog (approved):

ArnaudWeyts (approved):

joaoinacio (approved):

stivni:
Are participant and product mutually exclusive? Probably those fields will be filled in based on the type? I suggest we try to make a structure that represents that in that case...

joaoinacio:
> Are participant and product mutually exclusive? Probably those fields will be filled in based on the type? I suggest we try to make a structure that represents that in that case...

No, not really. a `time_tracking_product` for example inherits the participant from the timetracking.

ketlane:
The participant is the person who produced the action (i.e. added an external order, tracked some time). It's not mutually exclusive with the product.

PR #345, opened 2019-04-10
Add the WorkOrders.finalize documentation

JordyDeGroeve:
Do this according to the proposal #290

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3675,7 +3675,7 @@ Update a draft work order.
 
 ### workOrders.finalize [POST /workOrders.finalize]
 
-Finalize a draft work order. This will also generate the work order pdf.
+Finalize a draft work order. This will also generate the work order PDF.
 
 + Request (application/json;charset=utf-8)
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -3650,6 +3650,17 @@ Add a draft work order.
             + type: `workOrder` (string)
             + id: `7c1d8672-f502-4333-9ea4-7a45add15115` (string)
 
+### workOrders.finalize [POST /workOrders.finalize]
+
+ Finalize a draft work order by generating the PDF.
+
+ + Request (application/json;charset=utf-8)
+
+     + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+
+ + Response 204
+
 # Group Other
 
 ## Migrating [/migrate]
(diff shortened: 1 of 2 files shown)


tiagobastos (approved):

apiary.apib
```diff
@@ -3650,6 +3650,17 @@ Add a draft work order.
             + type: `workOrder` (string)
             + id: `7c1d8672-f502-4333-9ea4-7a45add15115` (string)
 
+### workOrders.finalize [POST /workOrders.finalize]
+
+ Finalize a draft work order by generating the PDF.
```
WouterSioen: I'm not sure "by generating the PDF" is correct. Something like `which will generate the work order pdf` might make more sense?

WouterSioen (approved):

apiary.apib
JordyDeGroeve: Just copied the description from the proposal. I like`which will generate ..` better, but will wait for some more input on this before changing, if you don't mind.

dietrichm: Let's change this. Generating is a side effect, not the main thing of finalizing. The main thing is just the state of the WO that is being changed.

apiary.apib
```diff
@@ -3650,6 +3650,17 @@ Add a draft work order.
...
+
+ + Request (application/json;charset=utf-8)
+
+     + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+
+ + Response 204
```
dietrichm: B: excess indenting (there's a space in front of every line).

dietrichm (changes requested):

apiary.apib
JordyDeGroeve: Really? 😓 This should get picked up by the linter..

tiagobastos (approved):

src/09-workorders/work-orders.apib
```diff
@@ -113,3 +113,14 @@ Update a draft work order.
         + location (Address, optional, nullable)
 
 + Response 204
+
+### workOrders.finalize [POST /workOrders.finalize]
+
+Finalize a draft work order. This will also generate the work order pdf.
```
dietrichm: PP:
```suggestion
Finalize a draft work order. This will also generate the work order PDF.
```

dietrichm (approved):

dietrichm (dismissed):

JordyDeGroeve:
That one time I used the `commit suggestion` option was the one time I shouldn't.

WouterSioen (approved):

tiagobastos (approved):

dietrichm (approved):

dietrichm:
✅ Checked

PR #346, opened 2019-04-10
[CORE-2000] WorkOrder Update Materials

tiagobastos:


apiary.apib
```diff
@@ -3569,6 +3569,10 @@ Update a draft work order.
         + milestone_id: `e94c1363-3426-46a4-921d-fce03f39bbd5` (string, optional, nullable)
         + event_id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string, optional, nullable)
         + location (Address, optional, nullable)
+        + materials (array, optional, nullable)
```
WouterSioen: Q: I don't think we normally make arrays nullable. If you want to remove all materials, you just pass an empty list of materials, right?

JordyDeGroeve: So this should become an array type that's optional, no?

JordyDeGroeve (commented):
Let's adjust the `material` requirements.

apiary.apib
dietrichm: Yup, just optional.

WouterSioen (approved):

JordyDeGroeve (approved):

dietrichm (approved):

PR #349, opened 2019-04-15
PRO-1746: document for projectItems.updateInternalRate

joaoinacio:
Document projectItems endpoint for setting participant rates for timetracking projectitems (for all projects or a single one).

### DIFF
--- a/src/07-projects/project-items.apib
+++ b/src/07-projects/project-items.apib
@@ -121,3 +121,19 @@ Get details for a single project item.
                 + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
             + added_at: `2018-12-04T16:44:33+00:00` (string)
             + updated_at: `2018-12-05T16:44:33+00:00` (string)
+
+
+### projectItems.setParticipantRates [POST /projectItems.setParticipantRates]
+
+Set participant user rates for timetracking project items, optionally for a single project
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, optional)
+        + participant_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_cost (Money, nullable)
+        + billable_price (Money, nullable)
+
++ Response 204 (application/json;charset=utf-8)
--- a/src/07-projects/project-items.apib
+++ b/src/07-projects/project-items.apib
@@ -137,7 +137,7 @@ Update a project item's internal cost.
 ### projectItems.updateInternalRate [POST /projectItems.updateInternalRate]
 
 Update the internal rate for a participant for one project, or all projects.
-This does not update the global internal rate of the employee. See `users.updateRates`.
+This does not update the global internal rate of the employee. See `users.update`.
 
 + Request (application/json;charset=utf-8)
 
--- a/src/07-projects/project-items.apib
+++ b/src/07-projects/project-items.apib
@@ -125,15 +125,14 @@ Get details for a single project item.
 
 ### projectItems.updateInternalRate [POST /projectItems.updateInternalRate]
 
-Set participant user rates for timetracking project items, optionally for a single project
+Update the internal rate for a participant for one project, or all projects.
+This does not update the global internal rate of the employee. See `users.updateRates`.
 
 + Request (application/json;charset=utf-8)
 
     + Attributes (object)
-        + filter (object, optional)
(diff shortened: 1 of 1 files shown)


src/07-projects/project-items.apib
```diff
@@ -120,4 +120,20 @@ Get details for a single project item.
                 + type: `product` (string)
                 + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
             + added_at: `2018-12-04T16:44:33+00:00` (string)
-            + updated_at: `2018-12-05T16:44:33+00:00` (string)
\ No newline at end of file
+            + updated_at: `2018-12-05T16:44:33+00:00` (string)
+
+
+### projectItems.setParticipantRates [POST /projectItems.setParticipantRates]
```
ArnaudWeyts: should we call it `updateParticipantRates` for consistency? Since we're also using `updateCost`

src/07-projects/project-items.apib
```diff
@@ -120,4 +120,20 @@ Get details for a single project item.
...
+
+Set participant user rates for timetracking project items, optionally for a single project
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, optional)
```
adognameddog: Q: is it necessary to have this inside `filter`? Could be just `project_id` (but nullable of course).

src/07-projects/project-items.apib
```diff
@@ -120,4 +120,20 @@ Get details for a single project item.
...
+        + participant_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_cost (Money, nullable)
+        + billable_price (Money, nullable)
```
ArnaudWeyts: Not sure if billable_price should be in here? We only need this endpoint to update the internal cost so I'm not sure why this is needed

src/07-projects/project-items.apib
```diff
@@ -120,4 +120,20 @@ Get details for a single project item.
...
+
+Set participant user rates for timetracking project items, optionally for a single project
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + filter (object, optional)
+            + project_id: `082e6289-30c5-45ad-bcd0-190b02d21e81` (string, optional)
+        + participant_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_cost (Money, nullable)
```
adognameddog: PP: I would use `internal_hourly_rate`, since we use the word "rate" in the endpoint.

src/07-projects/project-items.apib
adognameddog: I would opt for `projectItems.setInternalRate`

adognameddog (dismissed):

src/07-projects/project-items.apib
joaoinacio: > Q: is it necessary to have this inside `filter`? Could be just `project_id` (but nullable of course).

Not certain. if we foresee any other possible filters in the future it would make sense, otherwise - if it only the project will be used - then we can drop the filter and use just project_id.

As this documentation isn't released, though, perhaps we can just use project_id for now

ArnaudWeyts (dismissed):

src/07-projects/project-items.apib
ketlane: I guess once you moved the project_id outside of filter, you forgot to delete the `filter` field

src/07-projects/project-items.apib
```diff
@@ -120,4 +120,20 @@ Get details for a single project item.
...
                 + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
             + added_at: `2018-12-04T16:44:33+00:00` (string)
-            + updated_at: `2018-12-05T16:44:33+00:00` (string)
\ No newline at end of file
+            + updated_at: `2018-12-05T16:44:33+00:00` (string)
+
+
+### projectItems.updateInternalRate [POST /projectItems.updateInternalRate]
+
+Set participant user rates for timetracking project items, optionally for a single project
```
ketlane: Q: does this endpoint also update the user's global internal rate? I think we should document this.

ArnaudWeyts (dismissed):

ketlane (dismissed):

adognameddog (dismissed):

ArnaudWeyts (dismissed):

adognameddog (approved):

ArnaudWeyts (approved):

ketlane (approved):

PR #350, opened 2019-04-15
PRO-1749: add users.update endpoint documentation

joaoinacio:
⚠️ DO NOT MERGE - `internal_rate` property is part of "profit on projects" feature flag, not yet publicly released.

Document new endpoint `user.update`, which - for now - allows specifying a users's internal rate.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -626,15 +626,15 @@ Get details for a single user.
             + time_zone: `Europe/Brussels` (string)
 
 
-### users.update [POST /users.update]
+### users.updateInternalHourlyCost [POST /users.updateInternalHourlyCost]
 
 Update details for a single user.
 
 + Request (application/json;charset=utf-8)
 
     + Attributes (object)
-        + user_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
-        + internal_rate (object, nullable)
+        + id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_hourly_cost (object, nullable)
             + value (Money)
 
 + Response 204 (application/json;charset=utf-8)
--- a/apiary.apib
+++ b/apiary.apib
@@ -625,6 +625,22 @@ Get details for a single user.
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
 
+
+### users.updateRates [POST /users.updateRates]
+
+Update user rates used for timetracking.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + user_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_rate (object, nullable)
+            + value (Money)
+        + billable_price (object, nullable)
+            + value (Money)
+
++ Response 204 (application/json;charset=utf-8)
+
 ## Custom Fields [/customFieldDefinitions]
 
 Custom fields are used to add additional data/properties to entities within Teamleader.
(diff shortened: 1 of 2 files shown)


adognameddog (dismissed):
Linting, though

adognameddog (dismissed):

ArnaudWeyts:
Same comment here as in the PR involving project items, should we just make this one a generic `update` for which you can supply the property you can update?

ketlane:
> Same comment here as in the PR involving project items, should we just make this one a generic `update` for which you can supply the property you can update?

I think it's good that we specialize this endpoint: in the UI, we react with a custom pop-up if the user chooses to update one of these two fields, indicating that these fields probably should not stay together with the "Update employee" functionality, even in the UI. Furthermore, maybe we should add some more functionality in the future to this endpoint, such as the timeframe for which the update should apply.

Come to think about it, maybe we should already plan for this, by having the two rate fields as more than Money fields, i.e.

```
internal_rate:
  - value: Money
```

so that in the future we could add more stuff to this, like:

```
internal_rate:
  - value: Money
  - from: date (optional)
  - to: date (optional)
```

src/01-general/users.apib
```diff
@@ -120,3 +120,17 @@ Get details for a single user.
                     + tr-TR
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
+
+
+### users.updateRates [POST /users.updateRates]
+
+Update user rates used for timetracking
```
ketlane: I think we should indicate that there's an alternative projectItems.updateInternalRate endpoint that can only update some profit entries. Similarly, in the related endpoint, we should explain that we also have this endpoint that can set a default rate as well.

joaoinacio: > I think we should indicate that there's an alternative projectItems.updateInternalRate endpoint that can only update some profit entries. Similarly, in the related endpoint, we should explain that we also have this endpoint that can set a default rate as well.

However the documentation for projectItem endpoints is not actually available/released yet?

ArnaudWeyts (dismissed):

ArnaudWeyts (dismissed):

src/01-general/users.apib
ketlane: True, but I think we should hold this users.updateRates back as well (the concept of internal rate is only visible for users that have the profit feature flag for now). And by linking I don't mean creating a hyperlink, but just mentioning it should be enough for now.

apiary.apib
```diff
@@ -625,6 +625,22 @@ Get details for a single user.
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
 
+
+### users.updateRates [POST /users.updateRates]
```
ketlane: We need to keep this PR open until we fully release profit on projects: the internal rate is not visible unless you have the profit FF enabled.

ketlane (dismissed):

ArnaudWeyts (dismissed):

src/01-general/users.apib
```diff
@@ -120,3 +120,17 @@ Get details for a single user.
...
+
+
+### users.update [POST /users.update]
+
+Update details for a single user.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + user_id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
```
jenssegers:
```suggestion
        + id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
```

jenssegers:
After reading @ketlane's comment about passing additional parameters when updating the internal rate it makes more sense to have a dedicated endpoint to set the internal rate.

I'm guessing doing this will/might have side-effects? So in that way, it should definitely be its own endpoint.

joaoinacio:
closing in favor of https://github.com/teamleadercrm/api/pull/355

PR #355, opened 2019-04-16
document new users.updateInternalHourlyCost endpoint

joaoinacio:
⚠️ DO NOT MERGE - `iinternal_hourly_cost rate` property is part of "profit on projects" feature flag, not yet publicly released.

Document new endpoint which allows specifying a users's internal rate.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -628,7 +628,7 @@ Get details for a single user.
 
 ### users.updateInternalHourlyCost [POST /users.updateInternalHourlyCost]
 
-Update the internal hourly cost for a single user.
+Update the internal hourly cost for a single user. Only account administrators can set the internal hourly cost of a user.
 
 + Request (application/json;charset=utf-8)
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -625,6 +625,20 @@ Get details for a single user.
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
 
+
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -625,6 +625,20 @@ Get details for a single user.
...
+### users.updateInternalHourlyCost [POST /users.updateInternalHourlyCost]
+
+Update details for a single user.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `b5094b3f-bb7a-0391-b01b-e709773f3509` (string)
+        + internal_hourly_cost (object, nullable)
+            + value (Money)
```
ketlane: Not sure it's worth having value as a separate field inside internal_hourly_cost anymore. It was useful when the endpoint was more generic (i.e. updateRates). But now that it refers specifically to the internal hourly cost, I guess we don't need nested fields. You can easily just add a `from` and a `to` at the base level, along `internal_hourly_cost`

ketlane (dismissed):

apiary.apib
adognameddog: Precisely. Now that `value` is a bit over-engineered.

joaoinacio: indeed. updated.

adognameddog (dismissed):

ketlane (dismissed):

ArnaudWeyts (dismissed):

adognameddog (approved):

ketlane (approved):

PR #358, opened 2019-04-17
Add documentation for a workOrders.attachImage endpoint

WouterSioen:


apiary.apib
```diff
@@ -3673,6 +3673,23 @@ Update a draft work order.
 
 + Response 204
 
+### workOrders.attachImage [POST /workOrders.attachImage]
+
+Attach image to a draft work order.
```
JordyDeGroeve: PP: `Attach an image ..`

apiary.apib
```diff
@@ -3673,6 +3673,23 @@ Update a draft work order.
...
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+        + image: `data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAIAAACQkWg2AAAAAXNSR0IArs4c6QAAAVlpVFh0WE1MOmNvbS5hZG9iZS54bXAAAAAAADx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8AAAASUVORK5CYII=` (string, required) - PNG image, base64 encoded
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (object)
+            + type: `workOrderImage` (string)
```
JordyDeGroeve: Q: the type is not 100% decided yet, right?

WouterSioen: Not sure, this is what was in the original work order proposal

apiary.apib
```diff
@@ -3673,6 +3673,23 @@ Update a draft work order.
...
 + Response 204
 
+### workOrders.attachImage [POST /workOrders.attachImage]
+
+Attach an image to a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
```
WouterSioen: @dietrichm wouldn't it be clearer if this would be `workorder_id`?

apiary.apib
dietrichm: We discussed this yesterday and thought it should just remain a `fileId` just to avoid having to convert between WO and File UUIDs everywhere.

Only now got to updating the proposal. Sorry about that.

apiary.apib
dietrichm: Yes! Updated the proposal.

apiary.apib
WouterSioen: No worries, I somehow missed that decision

JordyDeGroeve (approved):
👍

tiagobastos (approved):

dietrichm (approved):

src/09-workorders/work-orders.apib
```diff
@@ -114,6 +114,23 @@ Update a draft work order.
...
 + Response 204
 
+### workOrders.attachImage [POST /workOrders.attachImage]
+
+Attach an image to a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + work_order_id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
```
jenssegers, line 124: Same comment here as https://github.com/teamleadercrm/api/pull/365, use `id`.

PR #359, opened 2019-04-17
history.list

psrpinto:
Our immediate goal is to build the Activity Stream (aka Tracking) on mobile, and we decided to first make this endpoint private, since there's lots of unknowns at the moment.

I took Stijn's [proposal](https://github.com/teamleadercrm/api-future/pull/186) and adapted it for our immediate needs. I have some doubts about some things, see comments on the commits.

### DIFF
--- a/src/01-general/history.apib
+++ b/src/01-general/history.apib
@@ -14,3 +14,25 @@ Retrieve a list of records about an item.
                          + contact
                          + company
                 + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string, required)
+
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string)
+                + title: `Marc was on a call with John from this company` (string)
+                + description: `An extended description` (string)
+                + recorded_at: `2017-05-09T11:25:11+00:00` (string)
+                + relates_to (object)
+                    + (object)
+                        + type (enum[string])
+                            + Members
+                                + contact
+                                + company
+                                + deal
+                                + invoice
+                                + meeting
+                                + call
+                                + timeTracking
+                         + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string)
--- a/src/01-general/history.apib
+++ b/src/01-general/history.apib
@@ -24,6 +24,37 @@ Retrieve a list of records about an item.
                 + title: `Marc was on a call with John from this company` (string)
                 + description: `An extended description` (string)
                 + recorded_at: `2017-05-09T11:25:11+00:00` (string)
+                + type (enum[string])
+                    + Members
+                        + call
+                        + client_form_filled_out
+                        + client_satisation_mail_sent_project
+                        + client_satisation_mail_sent_ticket
+                        + company
+                        + contact
+                        + creditnote
+                        + custom
+                        + deal
+                        + edit
+                        + email
+                        + event
+                        + external_interaction
+                        + file
+                        + gmail_email
+                        + invoice
+                        + linkedin
+                        + mail_sent
+                        + mailing
+                        + meeting
+                        + note
+                        + order
+                        + permanent_note
+                        + product
+                        + project
+                        + sale
+                        + ticket_closed
+                        + timetracking
+                        + todo
                 + relates_to (object)
                     + (object)
                         + type (enum[string])
--- a/src/01-general/history.apib
+++ b/src/01-general/history.apib
@@ -4,4 +4,13 @@
 
 Retrieve a list of records about an item.
 
++ Request (application/json;charset=utf-8)
 
+    + Attributes (object)
+        + filter (object)
+            + about (object, required)
+                + type (enum[string])
+                     + Members
+                         + contact
+                         + company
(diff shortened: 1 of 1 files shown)


src/01-general/history.apib
```diff
@@ -14,3 +14,25 @@ Retrieve a list of records about an item.
...
+                + recorded_at: `2017-05-09T11:25:11+00:00` (string)
+                + relates_to (object)
+                    + (object)
+                        + type (enum[string])
+                            + Members
+                                + contact
+                                + company
+                                + deal
+                                + invoice
+                                + meeting
```
psrpinto, line 66: `meeting` and `call` aren't resources in the API yet, they're both `event`s. However, the plan is to create a Calls API, at which point `event` will represent a `meeting`. We decided to keep using `meeting` here for clarity.

src/01-general/history.apib
```diff
@@ -24,6 +24,37 @@ Retrieve a list of records about an item.
...
+                        + meeting
+                        + note
+                        + order
+                        + permanent_note
+                        + product
+                        + project
+                        + sale
+                        + ticket_closed
+                        + timetracking
+                        + todo
```
psrpinto, line 57: This is the list of types that exist today, which obviously makes no sense in the API. Stijn proposes a different list, but we'd have to agree on it. Not sure what to do.

src/01-general/history.apib
```diff
@@ -0,0 +1,69 @@
...
+                        + mailing
+                        + meeting
+                        + note
+                        + order
+                        + permanent_note
+                        + product
+                        + project
+                        + sale
+                        + ticket_closed
+                        + timetracking
```
JorenSaey, line 56: no camelcase

src/01-general/history.apib
```diff
@@ -0,0 +1,69 @@
...
+                    + (object)
+                        + type (enum[string])
+                            + Members
+                                + contact
+                                + company
+                                + deal
+                                + invoice
+                                + meeting
+                                + call
+                                + timeTracking
```
JorenSaey, line 68: camelcase: should be consistent I think

src/01-general/history.apib
```diff
@@ -0,0 +1,69 @@
...
+                        + note
+                        + order
+                        + permanent_note
+                        + product
+                        + project
+                        + sale
+                        + ticket_closed
+                        + timetracking
+                        + todo
+                + relates_to (object)
```
psrpinto, line 58: In Stijn's proposal this is called `about`, but this isn't the same `about` as in the `filter`, so we decided to rename it, otherwise it would be confusing.

The `about` in the filter is, for example, the `contact` we want to see the history of. This `relates_to` is, for example, the `call` that was made to that `contact`.

src/01-general/history.apib
```diff
@@ -0,0 +1,69 @@
...
+                    + (object)
+                        + type (enum[string])
+                            + Members
+                                + contact
+                                + company
+                                + deal
```
psrpinto, line 64: What should we do when the related resource was deleted? Should we just set `relates_to` to `null`? Should `relates_to` have an extra field `deleted`? Another option?

src/01-general/history.apib
```diff
@@ -4,4 +4,13 @@
 
 Retrieve a list of records about an item.
 
++ Request (application/json;charset=utf-8)
 
+    + Attributes (object)
+        + filter (object)
+            + about (object, required)
+                + type (enum[string])
```
dietrichm: PP:
```suggestion
                + type (enum[string], required)
```

src/01-general/history.apib
```diff
@@ -14,3 +14,25 @@ Retrieve a list of records about an item.
...
++ Response 200 (application/json;charset=utf-8)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string)
+                + title: `Marc was on a call with John from this company` (string)
+                + description: `An extended description` (string)
+                + recorded_at: `2017-05-09T11:25:11+00:00` (string)
+                + relates_to (object)
```
dietrichm, line 58: B: 
```suggestion
                + relates_to (array)
```

(Unless it would be a single entity, then the object beneath it doesn't make sense.)

However I'm also not sure about the naming. There is also still an [open proposal](https://github.com/teamleadercrm/api/pull/265) on adding `relates_to` to time tracking and timers, where it represents the related entities of a time tracking/timer's subject. I.e. a project ID that relates to the milestone, and the customer of the project. So, a list of entities that we can deduce later.

Here it's actually kind of an actor, or entity that was acted on, right?

src/01-general/history.apib
dietrichm: I think the `id` should be `null` to make clear that you can no longer reference the entity, eventually with an added `(is_)deleted` property.

dietrichm (changes requested):

jenssegers:
@regularjack Can we move this to a meeting? Discussing this in a PR will not be efficient imo.

psrpinto:
@jenssegers Scheduled a meeting

psrpinto:
Ok, change of plans, or better, going back to the original plan. Since this is an "internal" endpoint, we won't put too much effort into doing it correctly. That was actually the idea behind making it internal. We can later iterate on it until we get something up to our API standards. 

Sorry for the hassle :|

PR #360, opened 2019-04-17
[CORE-2084] WorkOrders Info | Add distance

tiagobastos:


WouterSioen (approved):

JordyDeGroeve (approved):

src/09-workorders/work-orders.apib
```diff
@@ -37,6 +37,7 @@ Get details for a single work order.
                 + type: `event` (string)
                 + id: `9fc14045-5c6c-4ba8-8672-42ea3f26aa63` (string)
             + location (Address)
+            + distance (Distance, optional)
```
dietrichm: B:
```suggestion
            + distance (Distance)
```

dietrichm (dismissed):

dietrichm (approved):

JordyDeGroeve (approved):

JordyDeGroeve:
Q: so what do we return if there is no distance given?

mrtus:
@JordyDeGroeve @dietrichm maybe make it `nullable`?

WouterSioen:
I also think this should be nullable

tiagobastos:
In the proposal the distance is optional in the `.draft` and optional & nullable in the `.update`.
This being the `.info` is ok like this, right? 
It's consistent with all the others.

WouterSioen:
If it can be `null` or a `Distance` (since we support not passing it through draft and making it `null` through update), we should probably also reflect that in the `info` endpoint? It's not like we're going to make up some random data if there is no distance, right?

JordyDeGroeve (dismissed):
Also, as @tiago and @WouterSioen noticed, we should not allow floats to be sent as the value. This should be updated too.

WouterSioen (approved):

JordyDeGroeve (approved):

dietrichm (approved):

PR #365, opened 2019-04-23
Add proposal for the workOrders.deleteImage endpoint

WouterSioen:


JordyDeGroeve (approved):

tiagobastos (approved):

dietrichm (approved):

src/09-workorders/work-orders.apib
```diff
@@ -131,6 +131,18 @@ Attach an image to a draft work order.
...
             + id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string)
 
+### workOrders.deleteImage [POST /workOrders.deleteImage]
+
+Delete an attached image from a draft work order.
+
++ Request (application/json;charset=utf-8)
+
+    + Attributes (object)
+       + work_order_id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
```
jenssegers: I think if it's an action on the main resource we should use `id` instead of `work_order_id`. Similar to `linkToCompany` on contacts.

jenssegers (changes requested):

src/09-workorders/work-orders.apib
JordyDeGroeve: This feels more like a **PP** than a **B** to be honest.
We discussed this a week ago in the proposal. https://github.com/teamleadercrm/api/pull/290#discussion_r276205853

jenssegers: Since it would break consistency in the API I considered it to be blocking.

WouterSioen: We can still change this. @dietrichm what do you think, do I change this first before mobile has to do too many changes here?

dietrichm: The subject of the endpoint should always be passed using `id` indeed. We missed this when renaming the property last week. Let's change this back.

WouterSioen: Will do!a

tiagobastos (approved):

dietrichm (approved):

JordyDeGroeve (approved):

PR #367, opened 2019-04-23
Add a filter status on users.list

mrtus:
To be able to implement the inviting of users and the future of unique email addresses. We first need to be able to filter on all users to find if an email is already used or not. This will be the first step in doing so.

In a follow-up a filter `email` will be added.

Not sure yet how we would show the status in the response though.

Do we show the status as an enum:
```
{
    "status": "active|deactivated|...inactive"
}
```
Or do we show it as a boolean:
```
{
    "active": true
}
```

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -589,6 +589,10 @@ Get a list of all users.
                 + telephones (array[Telephone])
                 + language: `nl` (string)
                 + function: `Sales` (string)
+                + status: `active` (enum[string])
+                    + Members
+                        + active
+                        + deactivated
 
 ### users.info [GET /users.info]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -555,10 +555,11 @@ Get a list of all users.
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `John` (string, optional) - Filters on first name, last name and email
-            + status: `active`, `deactivated` (enum[string], optional) - Filters on status
-                + Members
-                    + active - Filters on active users
-                    + deactivated - Filters on deactivated users
+            + status: `active`, `deactivated` (array[string], optional) - Filters on status
+                + (enum[string])
+                    + Members
+                        + active - Filters on active users
+                        + deactivated - Filters on deactivated users
         + sort (array, optional)
             + (object)
                 + field (enum[string], required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -555,6 +555,10 @@ Get a list of all users.
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `John` (string, optional) - Filters on first name, last name and email
+            + status: `all` (enum[string], optional) - Filters on status
+                + Members
+                    + all - Filters on all active and deactivated users
+                    + deactivated - Filters only on deactivated users
         + sort (array, optional)
             + (object)
                 + field (enum[string], required)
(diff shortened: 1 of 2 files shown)


src/01-general/users.apib
```diff
@@ -51,6 +51,10 @@ Get a list of all users.
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `John` (string, optional) - Filters on first name, last name and email
+            + status: `all` (enum[string], optional) - Filters on status
```
dcsg: PP: I would allow an array of status if you want to filter by more than one. We did the same in the `invoice.list` https://github.com/teamleadercrm/api/blob/master/src/05-invoicing/invoices.apib#L16

src/01-general/users.apib
```diff
@@ -51,6 +51,10 @@ Get a list of all users.
...
+                + Members
+                    + all - Filters on all active and deactivated users
```
dcsg: PP: I believe `all` does not make much sense because IMO filtering all is the same as not using this filter at all.

mrtus: Currently, we do not return deactivated users, and since it doesn't make sense to always return deactivated users we decided to do it like this.

dcsg: Indeed, good point! Anyway, I would remove the `all` and allow an array of status to be filtered on.

src/01-general/users.apib
mrtus: Not sure it makes a lot of sense here though. I understand that for invoices you would have a lot more complex filters. But on users, I am not so sure since we never had the requests to do so.  (We currently have only two statuses too)

I'd say that we can always make a change and a BC thing if we actually need it.

Thoughts?

dcsg: I would prefer an array of status, I believe it would be nice if we implement it in the same way in every place. But let's wait to see what @stivni and @jenssegers think about this.

stivni: I would only implement it where it makes sense...

src/01-general/users.apib
stivni: the `all` case is actually a good reason do make the filter an array...

src/01-general/users.apib
stivni: given my comment about `all`, I do think it makes sense here

src/01-general/users.apib
mrtus: Yes, I agree, that argument convinced me to change the implementation.

dcsg (dismissed):

jenssegers:
I prefer the enum proposal instead of the boolean way. We might need to think about that enum options to return. Eg. do we return `deactivated` or `inactive`? Our UI displays them as "inactive users".

If we add it as a filter you should also add it to the response in the same way.

mrtus:
@jenssegers `deactivated` is the correct term. `Inactive` is a state where users didn't show activity for a certain period of time.

> Our UI displays them as "inactive users". 

Not anymore 😉

rathesDot (dismissed):

jenssegers:
Fair enough. Then I would just add the status to the response and you're good to 🚀

jenssegers (dismissed):
Also the return property in responses so that it is clear what we're filtering on.

dcsg:
Sorry for the push force, but Git Kraken changed my emails in my local repos :/

dcsg (approved):

jenssegers (approved):

rathesDot (approved):

apiary.apib
```diff
@@ -555,6 +555,11 @@ Get a list of all users.
         + filter (object, optional)
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + term: `John` (string, optional) - Filters on first name, last name and email
+            + status: `active`, `deactivated` (array[string], optional) - Filters on status
+                + (enum[string])
+                    + Members
+                        + active - Filters on active users
+                        + deactivated - Filters on deactivated users
```
psrpinto, line 562: PP: Feels a bit weird to me that we use the past tense for `deactivated` but the present tense for `active`.

mrtus: I think it makes sense since the user is still "active", the user has been "deactivated".

PR #370, opened 2019-04-24
Allow signing WorkOrders during finalization

WouterSioen:
Since we don't want signed work orders to still change, we need to sign
during the finalization step.


tiagobastos (approved):

JordyDeGroeve (approved):
Sing away! 🎤 🎶

jenssegers (approved):

dietrichm (approved):

PR #375, opened 2019-04-26
Remove charset from requests and responses

tiagobastos:
We removed the `charset=utf-8` from content-type in requests and responses.

Let's make this consistent in this feature branch.


tiagobastos:
Apparently, we already did this but the PR wasn't merged yet 😅

PR #378, opened 2019-04-29
[TOOL-174] Add `status` to `/users.info`

dcsg:
The `/users.list` now allows the integrators to filter by `active` and `deactivated` users, with this change now makes sense to display the user status in this endpoint and allow deactivated users to be returned.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -633,6 +633,10 @@ Get details for a single user.
                     + tr-TR
             + function: `Sales` (string)
             + time_zone: `Europe/Brussels` (string)
+            + status: `active` (enum[string])
+                + Members
+                    + active
+                    + deactivated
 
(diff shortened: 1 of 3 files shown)


mrtus (dismissed):

rathesDot (dismissed):

rathesDot:
Just wondering, is this a breaking change? Because with this we would return more users that before on the exact same call. (404 error vs. one result)

dcsg:
@rathesDot I don't think it is a breaking change since we are allowing more users to be returned.

jenssegers (dismissed):

mrtus (approved):

jenssegers (approved):

rathesDot (approved):

PR #382, opened 2019-05-02
WorkOrders List | Add Filter

tiagobastos:


apiary.apib
```diff
@@ -3566,6 +3566,18 @@ Get a list of work orders.
 + Request (application/json)
 
     + Attributes (object)
+        + filter (object, optional)
+            + customer (object, optional)
```
JordyDeGroeve: B: don't include the filter on customer yet.

JordyDeGroeve (dismissed):
Meant to block, sorry

apiary.apib
tiagobastos: 🤦‍♂

JordyDeGroeve (commented):
Q: do we only allow 1 state to be passed?

tiagobastos:
I think so yes. 
The state field is enumerating possible members.

apiary.apib
```diff
@@ -3566,6 +3566,12 @@ Get a list of work orders.
 + Request (application/json)
 
     + Attributes (object)
+        + filter (object, optional)
+            + state (enum[string], optional)
```
WouterSioen: This should be array of state, right? To allow multiple states at once?

JordyDeGroeve: Let's go for the multiple states. 👍

dietrichm (dismissed):
B: we'll be allowing filters on multiple states.

apiary.apib
```diff
@@ -3567,11 +3567,12 @@ Get a list of work orders.
 
     + Attributes (object)
         + filter (object, optional)
-            + state (enum[string], optional)
-                + Members
-                    + draft
-                    + finalized
-                    + sent
+            + state: `draft`, `sent` (array[string], optional) - The states you want to filter by
+                + (enum[string], required)
```
dietrichm: Strong PP: you can use
```
            + state (array[enum[string]], optional)
                + Members
                    + draft
                    + finalized
                    + sent
```
The example enum values tend to mess up the preview in the live docs. Also the remark is not strictly needed (pretty obvious what it does).

JordyDeGroeve (approved):

dietrichm (approved):

WouterSioen (approved):

PR #384, opened 2019-05-02
PRO-1839: ProjectItems.info: document subject (also work_type, activity_type)

joaoinacio:
~~### ⚠️ Rebase to master~~

Implementation PR: https://github.com/teamleadercrm/core/pull/9662
E2E Tests: https://github.com/teamleadercrm/api-tests/pull/663

Document new `subject` property in projectItems.info (which is the call/meeting/task id)
Also add missing work_type and activity_type (already in projectItems.list)


ketlane (approved):

adognameddog (dismissed):

src/07-projects/project-items.apib
```diff
@@ -128,6 +128,19 @@ Get details for a single project item.
...
                 + type: `product` (string)
                 + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
+            + work_type (object, nullable)
+                + type: `workType` (string)
+                + id: `206bf39b-0161-0058-a015-e530082b0814` (string)
+            + activity_type (object, nullable)
+                + type: `activityType` (string)
+                + id: `ec907500-37d0-0245-8b16-2dbedd0a8ee0` (string)
+            + subject (object, nullable)
+                + type: `call` (enum[string])
```
ketlane: The type should be 'event' here, if we're using UUIDs

src/07-projects/project-items.apib
```diff
@@ -128,6 +128,19 @@ Get details for a single project item.
             + product (object, nullable)
                 + type: `product` (string)
                 + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
+            + work_type (object, nullable)
+                + type: `workType` (string)
+                + id: `206bf39b-0161-0058-a015-e530082b0814` (string)
+            + activity_type (object, nullable)
+                + type: `activityType` (string)
+                + id: `ec907500-37d0-0245-8b16-2dbedd0a8ee0` (string)
+            + subject (object, nullable)
```
ketlane, line 137: Q: Wondering if this would be better called 'event' instead of 'subject'

ketlane (dismissed):
The type of the `subject` field should be `event`.

src/07-projects/project-items.apib
joaoinacio: > Q: Wondering if this would be better called 'event' instead of 'subject'

don't think this applies any longer, as we're returning the info using the same "structure" as TimeTrackings right?

ArnaudWeyts (approved):

ketlane (approved):

adognameddog (approved):

src/07-projects/project-items.apib
```diff
@@ -128,6 +128,22 @@ Get details for a single project item.
...
+                + type: `task` (enum[string])
+                    + Members
+                        + company
+                        + contact
+                        + event
```
jenssegers, line 142: @regularjack What's the status on events/todos vs. calls/tasks/meetings? Should we be using events and tasks together or not?

psrpinto: Here's some context:

We implemented the Events API to be about `meetings`/`calls`/`tasks`. We want to move away from that so that Events API becomes only about `meetings`.

This means there will be both a Tasks API and a Calls API. Tasks API already exists, Calls API will exist soon™.

Todo is not a concept that exists in our API. What was previously referred to as Todo should now be referred to as Task. There are still cases where a resource returns a `todo_id` but those should be changed to `task_id` (while keeping `todo_id` for BC, of course).

So, in this particular case, and considering Calls API does not exist yet (so there is no concept of `call` yet), I believe that list is correct, as long as `event` is only used for Meetings and Calls, and `task` is used for Tasks.

joaoinacio: > I believe that list is correct, as long as `event` is only used for Meetings and Calls, and `task` is used for Tasks.

Yes, that's what we are returning for subject.

psrpinto (approved):

PR #385, opened 2019-05-03
Fix array of strings

tiagobastos:
Is this a bug? Do we want this? 
- Removing the `string` at the end of the array
- Removing the values that look like duplicated values

Transforms this:

image

Into this:

image


dcsg (approved):

luisaterrabrito (approved):

dietrichm (approved):

dietrichm:
✅ Checked

PR #387, opened 2019-05-06
Follow up | Nullable Customer

tiagobastos:
While preparing a WorkOrder from a meeting we allow to create one without an associated customer so we should document this scenario in the info and list endpoints.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3598,7 +3598,7 @@ Get a list of work orders.
                         + draft
                         + finalized
                         + sent
-                + customer (object)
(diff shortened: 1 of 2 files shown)


WouterSioen (approved):

JordyDeGroeve (approved):

dietrichm (approved):

dietrichm:
✅ Checked

This doesn't provide visual changes in the generated docs afaik though.

tiagobastos:
I think only on the json schema.

PR #389, opened 2019-05-07
Allow to filter work orders on starts_at

WouterSioen:


JordyDeGroeve (approved):
Do we want to use the same value for both filters? (not that it's bugging me, but feels unnatural)

tiagobastos (approved):

src/09-workorders/work-orders.apib
```diff
@@ -13,6 +13,8 @@ Get a list of work orders.
                     + draft
                     + finalized
                     + sent
+            + started_after: `2017-03-26T10:01:49+00:00` (string, optional)
```
jenssegers: Would suggest using `starts_after` and `starts_before`. We wanted to have `starts_at` in the work order to not make the suggestion that work orders are either in the past or future. This filter would also suggest that it's mainly for fetching past work orders while it could also be used to fetch future work orders right?

jenssegers (changes requested):
See Q above ☝️

jenssegers (approved):

tiagobastos (approved):

PR #392, opened 2019-05-08
timeTracking.list returns results in descending order

psrpinto:
Results are returned in descending order, API docs said ascending. This fixes it.


JorenSaey (approved):

mikeverf (approved):

stivni (dismissed):
this is a breaking change!
has it been implemented like that?
is this really necessary?

psrpinto:
@stivni Indeed, this is fixing a bug in the docs :)

r1dotl (approved):

PR #397, opened 2019-05-09
[CAL-1391] Add milestone_id and deal_id to /tasks.create and /tasks.update

r1dotl:
**TICKET: https://teamleader.atlassian.net/browse/CAL-1391**

As an API consumer, I want to be able to link a Milestone or a Deal to a Task, for that to be possible we need to add an optional milestone_id and deal_id to /tasks.create and /tasks.update

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -125,8 +125,8 @@ Update a task.
         + description (string, optional)
         + due_at: `2016-02-04T16:44:33+00:00` (string, optional)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
-        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
-        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional, nullable)
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional, nullable)
         + estimated_duration (object, optional)
             + unit: `min` (enum[string], required)
                 + Members
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -92,6 +92,8 @@ Create a new task.
         + description (string, required)
         + due_at: `2016-02-04T16:44:33+00:00` (string, required)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, required)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
         + estimated_duration (object, optional)
             + unit: `min` (enum[string], required)
                 + Members
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -124,6 +124,8 @@ Update a task.
         + description (string, optional)
         + due_at: `2016-02-04T16:44:33+00:00` (string, optional)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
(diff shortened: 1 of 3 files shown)


jenssegers:
Don't forget the changelog!

mikeverf (dismissed):

r1dotl:
> Don't forget the changelog!

Didn't know about it, but already added it!
Thanks.

psrpinto (dismissed):

mikeverf (dismissed):

JorenSaey (dismissed):

jenssegers (dismissed):

src/changes-backwards-compatible.apib
```diff
@@ -6,23 +6,24 @@ We list all backwards-compatible additions here. These are currently available i
...
 
+- In order to link milestones and deals to a task, we added `milestone_id` and `deal_id` as optional parameters to `tasks.create` and `tasks.update`.
 - We added `sent` as a boolean to `invoices.info` and `invoices.list`
 - We added an optional filter on `ids` to `invoices.list`.
 
 #### April 2019
 
 - In `users.info`, we added the `status` property to the response since it is now possible to retrieve deactivated users.
 
-- In `users.list`, you can now also request deactivated users, by using a new filter on `status`.
```
stivni: 😞 your IDE is set to autoremove trailing spaces from code files, but in markdown that doesn't make sense, because that's a soft enter... could you revert this please?

stivni:
I'm unsure about this one...

I want to avoid that our APIs become long lists of (top level) properties... if we would be able to link a task to 7 other things, we would have 7 top level properties more... that's not a very scalable solution in my opinion... This also makes sideloading less usable, because for every linked item, we would have to add an extra argument to the `include=` parameter.

Not against the proposal completely and also something we can change over time, but thinking out loud to see what other people think about this... cc @jenssegers @regularjack

psrpinto:
@stivni 

> I want to avoid that our APIs become long lists of (top level) properties... if we would be able to link a task to 7 other things, we would have 7 top level properties more... 

Do you mean we would instead have something like the following?

```
+ links (array)
    + (object)
        + type (enum[string])
             + Members
                 + deal
                 + milestone
        + id: `c9258836-f9a5-40cb-aa2a-d55c22991b93` (string)
```

The disadvantage of this is that we then would have to enforce that only one `deal` is present in the `links` property. Having a `deal_id` is more explicit in that sense.


> if we would be able to link a task to 7 other things

The thing is, you **are** able to link the Task to a million other things, that sounds like a problem that would need to be solved at the product level.

stivni:
> The disadvantage of this is that we then would have to enforce that only one deal is present in the links property.

That's a good argument for this proposal. In the current conditions that makes sense...

src/08-tasks/tasks.apib
```diff
@@ -122,6 +124,8 @@ Update a task.
         + description (string, optional)
         + due_at: `2016-02-04T16:44:33+00:00` (string, optional)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
```
stivni: B: should be nullable too, right? To be able to clear it?

src/08-tasks/tasks.apib
```diff
@@ -122,6 +124,8 @@ Update a task.
...
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
```
stivni: B: should be nullable too, right? To be able to clear it?

psrpinto (dismissed):

psrpinto:
This should be good to go now. Could you re-review? @stivni @jenssegers @mikeverf @JorenSaey

mikeverf (dismissed):

JorenSaey (dismissed):

jenssegers (dismissed):

psrpinto (approved):

psrpinto:
Had to merge master to fix conflicts.

JorenSaey (approved):

mikeverf (approved):

PR #398, opened 2019-05-17
Add ability to have WorkOrder materials without product ID

dietrichm:
We either require a product ID *or* name and price. Either way the name and price can be passed as well as of now.

This is required because in the web version you could already draft a WorkOrder from a meeting having a material that is not an existing product (thus without ID) plus you can alter its name and price in the same form.

We need to support these scenarios in the API and mobile app.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3677,7 +3677,7 @@ Get details for a single work order.
                         + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
                     + quantity: 3 (number)
                     + name: `1/2" pipe` (string)
-                    + price: 30.20 (number)
+                    + unit_price (Money)
             + images (array) - Only included with request parameter `includes=images`.
                 + (object)
                     + id: `a856642a-a476-4b69-8e42-e7d1badb2ce9` (string)
@@ -3719,10 +3719,10 @@ Add a draft work order.
                    + Properties
                        + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
                        + name: `1/2" pipe` (string, optional)
-                       + price: 30.20 (number, optional)
+                       + unit_price (Money, optional)
                    + Properties
                        + name: `1/2" pipe` (string, required)
-                       + price: 30.20 (number, required)
+                       + unit_price (Money, required)
 
 + Response 201 (application/json)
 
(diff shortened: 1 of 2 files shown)


WouterSioen (approved):

tiagobastos (approved):

tiagobastos (approved):
✅ No linting errors. Creates the apiary file with html version correctly displaying changes.

JordyDeGroeve (approved):

stivni (approved):

apiary.apib
```diff
@@ -3672,10 +3672,12 @@ Get details for a single work order.
                     + description: `Replacing pipework` (string)
             + materials (array)
                 + (object)
-                    + product (object)
+                    + product (object, nullable)
                         + type: `product` (string)
                         + id: `e2314517-3cab-4aa9-8471-450e73449040` (string)
                     + quantity: 3 (number)
+                    + name: `1/2" pipe` (string)
+                    + price: 30.20 (number)
```
jenssegers: Money?

apiary.apib
```diff
@@ -3712,8 +3714,15 @@ Add a draft work order.
                + description: `Replacing pipework` (string, optional)
        + materials (array, optional)
            + (object)
-               + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
                + quantity: 3 (number, required)
+               + One Of
+                   + Properties
+                       + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
+                       + name: `1/2" pipe` (string, optional)
+                       + price: 30.20 (number, optional)
```
jenssegers: Money here too

apiary.apib
jenssegers: Also; should we use `unit_price` like we do invoicing? Because now there can be confusion if this is the unit price or the total price.

WouterSioen: We considered this, but since we're not accepting a currency, we didn't think it made sense here, right?

Completely agreeing with the unit price though.

jenssegers: Even if we don't accept it I think it's a good practice to be consistent in how we handle money values and currencies. I'm perfectly fine with throwing an exception if you try to pass a currency that is different from the department one. At least then the customer knows that it's not possible and we keep our API consistent.

tiagobastos (approved):

WouterSioen (approved):

JordyDeGroeve (approved):

JordyDeGroeve:
Apiary is created correctly, but it's hard to see the difference what is included in the `OR` groups.

jenssegers (approved):
🥇

PR #400, opened 2019-05-21
Add BC break that renames werkbonnen context to work_orders

WouterSioen:
This is consistent with other places in the api.

### DIFF
--- a/src/changes-backwards-incompatible.apib
+++ b/src/changes-backwards-incompatible.apib
@@ -3,6 +3,10 @@
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
 
+#### Latest
+
+- We renamed the `context` field for work orders on customFieldDefinitions from `werkbonnen` to `work_orders`
+
 #### 2019-03-13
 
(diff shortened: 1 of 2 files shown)


src/changes-backwards-incompatible.apib
```diff
@@ -3,6 +3,10 @@
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
 
+#### Latest
```
stivni, line 6: B : we should pick a new version number...

WouterSioen: What's the logic for upcoming version numbers? Which date do we pick?

tiagobastos:
Don't we use future API names branches for these changes?  
We had `2019-05-15` but don't know if we have a new one.

JordyDeGroeve (approved):
Should we start a release branch with all breaking changes?

src/changes-backwards-incompatible.apib
WouterSioen: Is it always the wednesday in the guild week?

WouterSioen:
Moved to https://github.com/teamleadercrm/api/pull/401

PR #402, opened 2019-05-21
[IN1823] Deal status and the triggers for events

EOSullivanBerlin:
Needed for the pipeline deals view.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1699,7 +1699,7 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
     + Attributes (object)
         + filter (object, optional)
             + ids: `53a7e592-a136-4bae-ae15-517befd3d75d`,`bb50af79-55cd-4be0-8306-e9626c70a90f` (array[string], optional)
-            + exclude_lost: true (boolean)
+            + exclude_lost: false (boolean)
         + page (Page, optional)
 
 + Response 200 (application/json;charset=utf-8)
--- a/apiary.apib
+++ b/apiary.apib
@@ -1708,6 +1708,17 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
             + (object)
                 + id: `21efc56e-1ba8-469d-926a-e89502591b47` (string)
                 + name: `New` (string)
+                + deal_status: `new` (enum[string])
+                    + Members
+                        + new
+                        + open
+                        + won
+                        + lost
+                + triggers: (array)
+                    + (object)
+                        + action: `create_event` (enum[string])
+                            + Members
+                                + create_event
 
 ## Deal Sources [/dealSources]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -1709,13 +1709,7 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
             + (object)
                 + id: `21efc56e-1ba8-469d-926a-e89502591b47` (string)
                 + name: `New` (string)
-                + related_deal_status: `new` (enum[string])
-                    + Members
-                        + new
-                        + open
-                        + won
-                        + lost
-                    + actions: create_event (array[string])
+                + actions: create_event (array[string])
 
 ## Deal Sources [/dealSources]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -1699,7 +1699,7 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
     + Attributes (object)
         + filter (object, optional)
             + ids: `53a7e592-a136-4bae-ae15-517befd3d75d`,`bb50af79-55cd-4be0-8306-e9626c70a90f` (array[string], optional)
-            + exclude_lost: false (boolean)
+            + include_lost_phase: true (boolean)
         + page (Page, optional)
 
 + Response 200 (application/json;charset=utf-8)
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -1708,6 +1708,17 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
             + (object)
                 + id: `21efc56e-1ba8-469d-926a-e89502591b47` (string)
                 + name: `New` (string)
+                + deal_status: `new` (enum[string])
```
jenssegers: Could we get some more information why these fields are required?

During the initial deal API prototyping we discussed that we don't want new/open/won/lost to be part of deal phases, but as an additional status property on deal level. That way you can mark a deal as lost while it's in the "contact" deal phase or whatever.

Would it be possible to take this in scope while we're working on deals?

luisaterrabrito: why not just status?

EOSullivanBerlin: The triggers are for creating particular actions when a deal enters a phase. In this case we want to show a pop-up for creating a meeting. This behaviour already exists in the old overview. 

We need the deal_status so that we can for example hide the lost deal phase or show the 15 most recent won deals.

stivni: I think the consensus of our design meeting was that we can't change the way the phases work, because we can have phases after won or lost... so won & lost currently need to be phases... 
Yet, we still want to know which phases are for won, lost and new.
We call it "deal_status" because it describes the status of the deals within the phase, not of the phase themselves... If you have a better name suggestion, feel free to share...

WouterSioen: `related_deal_status`?

apiary.apib
```diff
@@ -1708,6 +1708,17 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
...
+                    + Members
+                        + new
+                        + open
+                        + won
+                        + lost
+                + triggers: (array)
```
jenssegers: PP:
```suggestion
                + actions: (array[string])
```

Actions can simply be an array of enums right?

apiary.apib
jenssegers: Can we replace this with a filter `excludeLost` that is only available for us? Feels like that solves the same problem and will be easier to keep hidden.

lowiebenoot: So what if we then need to do something special for the won phase (which might be the case)? This doesn't seem like a good solution to me.

apiary.apib
EOSullivanBerlin: Yes I think that would work fine.

src/03-deals/deal-phases.apib
```diff
@@ -22,3 +23,10 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
             + (object)
                 + id: `21efc56e-1ba8-469d-926a-e89502591b47` (string)
                 + name: `New` (string)
+                + related_deal_status: `new` (enum[string])
```
WouterSioen: B: the main apiary file should be re-build

WouterSioen (changes requested):

WouterSioen (dismissed):

WouterSioen (dismissed):

luisaterrabrito (dismissed):

jenssegers (dismissed):

apiary.apib
```diff
@@ -1699,6 +1699,7 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
     + Attributes (object)
         + filter (object, optional)
             + ids: `53a7e592-a136-4bae-ae15-517befd3d75d`,`bb50af79-55cd-4be0-8306-e9626c70a90f` (array[string], optional)
+            + exclude_lost: true (boolean)
```
```suggestion
            + exclude_lost: false (boolean)
```

Default value will be `false`.

WouterSioen: I'm seeing a double negation here. Isn't that something we try to avoid?

apiary.apib
Yarn-e: It feels weird to limit the filter to only `Lost`, what if we want to hide the `Won` phase or another type of phase ?

luisaterrabrito (approved):

WouterSioen (approved):

Yarn-e (approved):

apiary.apib
stivni: > It feels weird to limit the filter to only Lost, what if we want to hide the Won phase or another type of phase ?

yagni...

Although designs are changing towards a situation where we might want be able to identify the won phase... so we might have to swing back to the original idea... more info with @PieterReel

apiary.apib
stivni: @jenssegers we were just going to wrap them in objects so we could add meta-data to actions later on, but I guess that's a simple BC-layer so, we can do that later on, indeed 🤷‍♂

apiary.apib
stivni: @WouterSioen I indeed remember discussing the double negation part and agree with the comment... but we might need revert this commit, ask @PieterReel ...

stivni (approved):
approving with what we have here, but I think requirements are changing...

EOSullivanBerlin:
I heard there are some changes coming and also this will not be published so no merging.

jenssegers (approved):

PR #406, opened 2019-05-21
[Proposal] Scheduling Tasks

psrpinto:
This PR presents a proposal for scheduling Tasks without using the Events API. It focuses on how Tasks behave today, as opposed to the ideal scenario, which I believe was the intention behind the Events API. However, implementing that ideal scenario from day one is, I believe, extremely difficult, if not impossible. For concrete examples of these difficulties, please see the Background section below.

This proposal presents a solution that models the way Tasks behave today and addresses our current needs. I believe it also reduces complexity by an order of magnitude - compared to a solution that requires using both the Events API and the Tasks API - for ourselves who will implement it and consume it, and for external API consumers too.

Our current needs are:

- Schedule a Task, potentially multiple times
- Display a list of both scheduled and unscheduled Tasks
- Display a Task, including information about when that Task is scheduled

Please keep reading for more information into why I believe this to be the way to go. I think what follows provides undeniable evidence of the complexity that a two-API solution presents.

I'd love to hear your thoughts about this proposal.

## Background

The original vision for the Events API was that it would unify all *scheduled* resources - Meetings, Calls and Tasks - in a single API. In addition, given that Calls and Tasks aren't necessarily scheduled, the concept of a Todo was introduced, to represent either a Call or a Task. So we would have:

- Events API - Scheduled resources: Meetings, Scheduled Calls and Scheduled Tasks
- Todos API - (Unscheduled) Calls and (Unscheduled) Tasks

(Note that the Todos API was never implemented, more on that below.)

However, in the context of the Todos API, these resources currently behave differently, in multiple dimensions, for example:

- Calls must have an outcome once completed, Tasks must not
- Tasks must be linked to a WorkType, Calls must not
- Calls must have a link to a Customer, Tasks may have it
- Calls may have a link to both a Company and a Contact, Tasks can only be linked to one of either

Given these, and other, behavioural differences, it was decided to drop the concept of a Todo and implement both a Tasks API and a Calls API, tailored to how each resource behaves. The Tasks API has been implemented and the Calls API is planned for the near future. This would give us:

- Events API - Scheduled resources: Meetings, Scheduled Calls and Scheduled Tasks
- Tasks API - (Unscheduled) Tasks
- Calls API - (Unscheduled) Calls

However, having a resource "shift" between APIs once it is scheduled presents great challenges, mostly related to keeping state in sync between the Task or Call and the Event. For example:

- What should happen to the title or description of the Event when the title or description of the Task or Call changes? And vice-versa.
- What should happen to the Event when the Task or Call is marked as done?
- What should happen to the Event when the Task or Call is linked to a new Customer? Or Project, or Deal, or etc. And vice-versa.

To make matters worse, Tasks can be scheduled multiple times, which means for each Task there would potentially be multiple Events to keep in sync. We would also need to keep Events that relate to the same Task in sync between each other.

In addition, different types of scheduled resources also behave differently:

- Meetings may have a location, Calls and Tasks must not
- Meetings and Tasks may be recurring, Calls must not
- Meetings may have a list of participants, Calls and Tasks must not
- Calls must have a link to a Customer, Meetings must not, Tasks may have it


apiary.apib
```diff
@@ -3261,6 +3261,10 @@ Get a list of tasks.
                 + completed: false (boolean)
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
+                + scheduled_at: (array)
```
psrpinto: I'm not sure about the naming here. The `_at` part maybe implies it's a datetime when in fact it's an object with two datetimes. I couldn't come up with a better name, suggestions very welcome.

apiary.apib
```diff
@@ -3237,6 +3237,8 @@ Get a list of tasks.
             + ids: `9c475de2-1e99-0328-bc19-271b27e921d8`,`6325b4e5-c6a6-060b-b010-b81892e82fa7` (array[string], optional)
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
             + completed: true (boolean, optional)
+            + since: `2017-01-01T00:00:00+00:00` (string, optional) - Start of the period for which to return tasks. Takes in consideration both the `due_at` and `scheduled_at` fields.
+            + before: `2018-01-01T00:00:00+00:00` (string, optional) - Start of the period for which to return tasks. Takes in consideration both the `due_at` and `scheduled_at` fields.
```
psrpinto, line 3241: Not sure about the naming here either, suggestion welcome.

psrpinto:
Closing this PR for now and moving the discussion offline.

PR #408, opened 2019-05-22
workOrders endpoints - include document_template_id

JordyDeGroeve:
when a template is not set, we use the default. So a doc template is always known. That's why the returned id (in **.info**) is not nullable.


tiagobastos (approved):

tiagobastos:
✅ No linting errors. Create the expected apiary.html file.

WouterSioen (approved):
Is document_template how we're naming this? I thought (but I'm not entirely sure anymore) that we were going to name this templates? Let's double check this?

Looks fine otherwise.

src/09-workorders/work-orders.apib
```diff
@@ -94,6 +97,9 @@ Get details for a single work order.
             + department (object, nullable)
                 + type: `department` (string)
                 + id: `5e90eb0a-b502-4344-aa0f-3b8525af6186` (string)
+            + document_template (object)
```
dietrichm: B:
```suggestion
            + document_template (object, nullable)
```

dietrichm (changes requested):

dietrichm:
We agreed on `document_template` during grooming/sprint planning IIRC.

src/09-workorders/work-orders.apib
tiagobastos: So we don't return the default one, like it's said in the description?

apiary.apib
```diff
@@ -3653,6 +3653,9 @@ Get details for a single work order.
             + department (object, nullable)
                 + type: `department` (string)
                 + id: `5e90eb0a-b502-4344-aa0f-3b8525af6186` (string)
+            + document_template (object)
+                + type: `document_template` (string)
```
```suggestion
                + type: `documentTemplate` (string)
```

src/09-workorders/work-orders.apib
```diff
@@ -94,6 +94,9 @@ Get details for a single work order.
...
+                + type: `document_template` (string)
```
jenssegers:
```suggestion
                + type: `documentTemplate` (string)
```

jenssegers (dismissed):
Type reference should be camelCase

src/09-workorders/work-orders.apib
JordyDeGroeve: Like tiago said, we do know which template will be used. Why not return the default one?
(you mentioned that this is not the case for the department. Should we go for consistency or complete data?)

dietrichm: We actually do not do this for the department either. Why would we for the document template?

It only matters when we are finalizing. When it's not set on the WO, the default is being used.

src/09-workorders/work-orders.apib
JordyDeGroeve: Fixed!

apiary.apib
JordyDeGroeve: Fixed 👍

src/09-workorders/work-orders.apib
JordyDeGroeve: Made it nullable 😉

tiagobastos (approved):

WouterSioen (approved):

jenssegers (approved):

dietrichm (approved):

src/09-workorders/work-orders.apib
stivni: but a (draft) workorder always has a department and a template, no?
so why would you hide that from the API users?

stivni:
documentTemplates.list is at least an existing endpoint... so that name would be fine...
but feel free to give it a better name in the context of the workorder if you want... because the properties don't have to be named after the type, it sometimes even helps to make things more clear
(eg. `event { organiser: { type: user } }`)

stivni (approved):
I approve, but I think it would be better to return the value we would use for the template (and the department) but as that is not BC-breaking, we can do that as a follow up...

src/09-workorders/work-orders.apib
dietrichm: This can give two slightly weird scenarios though:

1. When upon drafting the default template is assigned in the background, and that default template is later changed to a different one, the WorkOrders that were not yet finalized will all be created with the old template.
1. When we return the current default template for every WO without one explicitly set, changing the default template will make workOrders.info suddenly return a different template. This can be confusing for an integrator that a value changed. Although this might be the best scenario in this case.

But I would also keep it as is and see if we need to change this later on.

PR #412, opened 2019-05-27
Provided and remaining budget on milestones.list and milestones.info [PRO-1935]

adognameddog:


src/changes-backwards-compatible.apib
```diff
@@ -10,6 +10,11 @@ We list all backwards-compatible additions here. These are currently available i
 - We added an optional filter on `ids` to `invoices.list`.
 - We added `creditNote.booked` and `product.added` to the available webhooks list.
 
+#### June 2019
```
stivni, line 15: I'm not sure how this fits the order of things here :)

jenssegers (dismissed):

src/07-projects/milestones.apib
```diff
@@ -89,6 +93,10 @@ Get details for a single milestone.
                     + time_and_materials
                     + fixed_price
                     + non_invoiceable
+            + `budget` (object) - This is only available for users that have access to budgets on projects.
+                + `allocated` (Money)
```
stivni: as discussed in person, I now think that "allocated" actually creates confusion around "is this allocated *to* the project, or allocated *within* the project"...

_provided_ budget might seem to be a good alternative...

jenssegers: Isn't "allocated" the estimated budget?

jenssegers: Also, if it's between allocated and provided I would choose allocated.

ArnaudWeyts (dismissed):

ketlane (approved):

src/changes-backwards-compatible.apib
ketlane: I guess we can reevaluate before merging

src/07-projects/milestones.apib
adognameddog: After some research we think "provided budget" for a milestone (so when you create a milestone you specify that budget) is better than "allocated". Allocated in this case means the budget you plan to spend, so haven't spent yet.

joaoinacio (approved):

jenssegers (approved):

ArnaudWeyts (approved):

stivni (approved):

PR #413, opened 2019-05-28
PROPOSAL: DocumentLayoutVariations.list

tiagobastos:
### ADDED
- `documentTemplateVariations.list` endpoint
- variations to the response of `documentTemplates.list`

### REMOVED 
- charset from the request and response

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -770,7 +770,7 @@ Get a list of all work types, sorted alphabetically (on their name).
 
 Get a list of all document templates.
 
-+ Request (application/json;charset=utf-8)
++ Request (application/json)
 
     + Attributes (object)
         + filter (object, optional)
@@ -785,7 +785,7 @@ Get a list of all document templates.
                     + timetracking_report
                     + workorder
 
-+ Response 200 (application/json;charset=utf-8)
++ Response 200 (application/json)
 
     + Attributes (object)
         + data (array)
@@ -821,7 +821,7 @@ Get a list of all document template variations.
             + ids: `cb8da52a-ce89-4bf6-8f7e-8ee6cb85e3b5`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], optional)
             + department_id: `a344c251-2494-0013-b433-ccee8e8435e5` (string, optional)
 
-+ Response 200 (application/json;charset=utf-8)
++ Response 200 (application/json)
 
     + Attributes (object)
         + data (array)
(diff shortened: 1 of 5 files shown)


WouterSioen (approved):

JordyDeGroeve (approved):
Q: what about the default layout? Is this seen as a variation?

stivni (changes requested):
I'm unsure we want to keep the concept of variations... 
This implies that there is a relationship between invoices - creditnotes - workorders etc...
While in reality, more people actually struggle with the fact that for every _variation_, they have to provide all kind of templates....

What is the original need you wanted to solve? Because I recall discussing this with @JordyDeGroeve yesterday, and this was not the discussed approach...

dietrichm:
@stivni I don't think this implies that there is a hard coded relation between the different types of entities. That is how it is in the UI but that is not reflected here. It only defines that a template can be in only one variation.

(Maybe in the future we will allow to have variations that _can_ contain templates for multiple entity types, but it doesn't have to be like that.)

We want to have a sensible name for every template returned in the response, so that a clear name (they provided) can be shown to the mobile user.

As the filters on the document template list endpoint are optional, just adding the name of the layout variation _will_ right now reveal that the variations are actually shared as well.

Of course if we totally want to ditch the concept of a variation it makes sense, but this is a decision product should make.

dietrichm (changes requested):
B: we agreed to call the endpoint `documentLayoutVariations.list`.

WouterSioen:
@dietrichm isn't that a bit weird since the templates itself has as type `documentTemplates`?

dietrichm:
We decided for layout variation because that's what it's called in the UI and afaik will be in the mobile app.

JordyDeGroeve (approved):

WouterSioen (approved):

jenssegers:
Was talking with @stivni yesterday about this and basically what you want is to show a drop-down with the available templates you can choose for a work order, invoice or whatever. Can't we simply add a read-only `name` property to the `documentTemplates.list` endpoint? That way you can just do a filter on `documentTemplates.list` and display the name in the dropdown.

Then you would get a dropdown with, for example:

 - default
 - variation with big logo
 - variation with small logo

Then there's no need to do any additional API calls or sideloading. To me, this also feels quite future proof because now we can still go any direction we want with document templates and we took a small step that won't block us in the future since it's read-only.

If in the future if we decide to let our customers create variations for a specific document without having to create an entire variation set for all your documents we can still do that. And if this turns out to be a bad decision, removing a read-only field in a new API version seems easier than deprecating an entire relationship.

tiagobastos:
Makes sense indeed. Closing then.

WouterSioen:
https://github.com/teamleadercrm/api/pull/418

PR #414, opened 2019-05-28
Makefile Reorder/Remove/Refactor Tasks 

tiagobastos:
This may be a particular preference of mine but, if someone forgets to make the lint it will fail in the CI so why not make that a prerequired tasks locally? 

### Changed
- `lint` is now a required task to build
- `build` task is now the default task for make command

### Removed
- `all` task

**Note:**
This change doesn't go against the CONTRIBUTING file because the
default command behavior still stands with the additional improvement of
linting before building or previewing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ node_modules: package.json
 lint: node_modules
 	docker run -it --init --rm -v $(PWD):/code -w /code node:9 npm run lint
 
-build: src node_modules
+build: src node_modules lint
 	docker run -it --init --rm -v $(PWD):/code -w /code node:9 npm run build
 
 preview: build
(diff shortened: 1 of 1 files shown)


WouterSioen (commented):
If you just run `make` on master, it already does both building and linting so I don’t really see the need for this tbh

psrpinto:
I prefer how it is currently since it's more explicit: `build` builds, with this proposal `build` builds and lints. I also agree with Wouter, I usually just do `make` which already does both.

jenssegers:
I remember the reason why `lint` was after `build` is because you first need the updated `apiary.apib` file before you can lint it 🤓

tiagobastos:
Valid remarks! 👍 
I was going for a `make preview` thas did build and lint 😅 
But if we can only lint the main apiary file that kind invalidates the proposal. 

And indeed this change would make it less explicit of what build task was doing.

PR #415, opened 2019-05-28
Work Orders List |  Filter on Customer

tiagobastos:


WouterSioen (approved):

JordyDeGroeve (approved):

jenssegers (approved):

JordyDeGroeve (approved):
✅ apiary gets created correctly

Q: any idea why we repeat the type `string` in the enum options?
Schermafbeelding 2019-05-29 om 10 39 02

This is also the case for deals.list, tasks.list, ..

tiagobastos:
Tried once to remove that but without success. 😞 

If the field is an array we do this: `array[enum[string]]` (check the filter on state) and it shows correctly. But this field is not an array so that `string` keeps showing up. 

@jens @dietrichm @wouterds any idea how to remove the extra `string`?

apiary.apib
```diff
@@ -3577,6 +3577,12 @@ Get a list of work orders.
             + starts_after: `2017-03-26T10:01:49+00:00` (string, optional)
             + starts_before: `2017-05-26T10:01:49+00:00` (string, optional)
             + term: `John Doe` (string, optional) - Filters on customer name, project name and milestone name
+            + customer (object, optional)
```
dietrichm, line 3577: PP: add it after `state` to keep the same order as the actual WO properties.

dietrichm (approved):
No idea how to fix the example at the moment. But I wouldn't put too much effort into fixing that right now.

jenssegers (approved):

JordyDeGroeve (approved):

PR #418, opened 2019-05-29
Return name of layout in the templates.list endpoint

WouterSioen:
This is necessary to visual display templates.

### DIFF
--- a/src/01-general/document-templates.apib
+++ b/src/01-general/document-templates.apib
@@ -38,3 +38,4 @@ Get a list of all document templates.
                         + timetracking_report
                         + workorder
                 + is_default: true (boolean)
+                + name: `new logo` (string)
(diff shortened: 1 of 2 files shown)


src/01-general/document-templates.apib
```diff
@@ -38,3 +38,4 @@ Get a list of all document templates.
                         + timetracking_report
                         + workorder
                 + is_default: true (boolean)
+                + name: `new logo` (string)
```
JordyDeGroeve: B: this should be a `nullable` string. We don't have a name for the default variations. We should return `null` in this case

JordyDeGroeve (changes requested):

JordyDeGroeve (approved):

JordyDeGroeve:
Q: should we add a comment in which case that we return `null`? Or is this not needed?

WouterSioen:
I don't feel like that's necessary?

jdrieghe (approved):

jenssegers (approved):

rathesDot (approved):

stivni (approved):

tiagobastos (approved):

PR #419, opened 2019-05-29
[CAL-1391] Return deal_id and milestone_id from tasks.info and tasks.list

r1dotl:
TICKET: https://teamleader.atlassian.net/browse/CAL-1391

Allow the tasks.list and tasks.info to return the milestone_id and deal_id.

If the milestone_id/deal_id don't exist it will return null, but the attribute will be always returned.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -277,6 +277,7 @@ You can find a (very) minimal PHP code example, showing how to start the OAuth 2
 
 A Teamleader PHP SDK is under development. In the meanwhile, we recommend using packages such [Guzzle](http://docs.guzzlephp.org/en/stable) or the [League OAuth2 client](http://oauth2-client.thephpleague.com).
 
+<a name="sideloading"></a>
 ## Sideloading
 
 Sideloading (or compound documents) allows you to retrieve relevant related items as part of a single request. For example, a deal has a reference to a lead which can either be a contact or a company:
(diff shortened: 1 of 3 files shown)


src/08-tasks/tasks.apib
```diff
@@ -48,6 +48,8 @@ Get a list of tasks.
                             + user
                             + team
                     + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
+                + milestone_id: `6fac0bf0-e803-424e-af67-76863a3d7d16` (string, nullable)
```
psrpinto: B: Should be a `milestone` object instead of `milestone_id`. Same for `deal` and same for `.info`.

psrpinto (changes requested):

src/changes-backwards-compatible.apib
```diff
@@ -5,7 +5,7 @@ We list all backwards-compatible additions here. These are currently available i
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
 #### Latest
-
+- `tasks.list` and `tasks.info` now returns the milestone and deal objects.
```
psrpinto:
```suggestion
- `tasks.list` and `tasks.info` now return the milestone and deal objects.
```

apiary.apib
```diff
@@ -3315,6 +3329,8 @@ Create a new task.
         + description (string, required)
         + due_at: `2016-02-04T16:44:33+00:00` (string, required)
         + work_type_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, required)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
```
psrpinto: B: These changes are not from this PR, right? Same for the whitespace changes.

psrpinto (changes requested):

mikeverf (approved):

psrpinto (approved):
I think it would have been better to figure out why the unwanted changes were here and act accordingly, instead of removing them through extra commits. Also, the whitespace changes are still there.

JorenSaey (approved):

jenssegers (approved):

stivni:
> I think it would have been better to figure out why the unwanted changes were here and act accordingly, instead of removing them through extra commits.

Indeed, because if these commits get to master, those lines will be removed if they are in master...

stivni (commented):
I wonder if we need to put this both as a top level property...
Have you considered other options, like a "links" section?
Not sure if that makes more sense here, but want to consider it for a second...

r1dotl:
This is already a mess, so i will close this one and open a new one clean.

r1dotl:
Creating a new one more clean.

PR #421, opened 2019-05-31
Added statuses for deal phases

EOSullivanBerlin:
I just pass along the phases as they are in our data
base.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1710,6 +1710,14 @@ Get a list of all phases a deal can go through, sorted by their order in the flo
                 + id: `21efc56e-1ba8-469d-926a-e89502591b47` (string)
                 + name: `New` (string)
                 + actions: create_event (array[string])
+                + status: `lost` (enum[string])
+                    + Members
+                        + new
+                        + won
+                        + lost
+                        + quotation_send
+                        + custom
+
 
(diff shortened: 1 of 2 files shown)


jenssegers:
Why do you need this @EOSullivanBerlin?

EOSullivanBerlin:
@jenssegers Pieter wants to create two areas at the bottom of the page called won and lost. He wants this because there could be up to 16 columns but sales people can move the deals to won from any stage. 
This is only possible if I can identify the won and lost phases.

stivni:
Then again, why should we treat those areas as phases? If they are not represented as phases?
Do we have some designs or wireframes for this?

jenssegers (changes requested):
I don't really agree that this is the way to go. We should try to keep deal phases independent from deal statuses.

EOSullivanBerlin:
This is why we want to be able to identify the phases: The won and lost sections at the bottom.  
We don't have wire frames as design and UX are super busy but the idea is to complete the MVP according to what Pieter Reel and Jeroen agreed. Then with the feed back and a proper study of what the users expect we can change and modify it accordingly. 
I would therefore not expect to be able to determine now what we will need from the API in the future and I would suggest that we don't publish it and just change it as we need it.

PR #423, opened 2019-06-03
work order materials: add purchase price

WouterSioen:
This can be manually passed to:

* overwrite the purchase price when adding a product as work order material
* set a purchase price when adding a material without a product attached (a one-time product)

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3691,6 +3691,7 @@ Get details for a single work order.
                     + quantity: 3 (number)
                     + name: `1/2" pipe` (string)
                     + unit_price (Money)
+                    + purchase_price (Money, nullable)
             + custom_fields (array[CustomField])
             + images (array) - Only included with request parameter `includes=images`.
                 + (object)
@@ -3730,6 +3731,7 @@ Add a draft work order.
        + materials (array, optional)
            + (object)
                + quantity: 3 (number, required)
+               + purchase_price (Money, optional)
                + One Of
                    + Properties
                        + product_id: `e2314517-3cab-4aa9-8471-450e73449040` (string, required)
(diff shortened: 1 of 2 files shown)


JordyDeGroeve (approved):

apiary.apib
```diff
@@ -3777,6 +3779,7 @@ Update a draft work order.
         + materials (array, optional)
             + (object)
                 + quantity: `2.2` (number, required)
+                + purchase_price (Money, optional)
```
JordyDeGroeve, line 3782: Q: This is only `optional` and not `optional, nullable` on purpose, right?

src/09-workorders/work-orders.apib
```diff
@@ -215,6 +217,7 @@ Update a draft work order.
         + materials (array, optional)
             + (object)
                 + quantity: `2.2` (number, required)
+                + purchase_price (Money, optional)
```
dietrichm, line 220: B: unless we decide it shouldn't be possible to unset the custom purchase price, but it makes sense in the overwritten purchase price scenario.
```suggestion
                + purchase_price (Money, optional, nullable)
```

dietrichm (changes requested):

src/09-workorders/work-orders.apib
WouterSioen: But it's a collection, we just overwrite the collection completely in updates, right? 🤔

apiary.apib
WouterSioen: It's a collection, we just overwrite the collection completely in updates, right? 🤔

src/09-workorders/work-orders.apib
dietrichm: Hmm in that sense you're right indeed.

dietrichm (approved):

tiagobastos (approved):

jenssegers (approved):

stivni (approved):

apiary.apib
```diff
@@ -3730,6 +3731,7 @@ Add a draft work order.
        + materials (array, optional)
            + (object)
                + quantity: 3 (number, required)
+               + purchase_price (Money, optional)
```
stivni, line 3734: PP: move to the bottom of this group

PR #425, opened 2019-06-04
Work hours: allow multiple assignees

WouterSioen:


apiary.apib
```diff
@@ -427,6 +427,10 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
 
+#### 2019-07-03
+
+- time added to work orders now has multiple users instead of one. You can pass a list of uuid's in the workOrders.draft and workOrders.update endpoints and will receive a list of user entities in workOrders.info.
```
JordyDeGroeve: PP/B: Start the sentence with a capital letter.
```suggestion
- Time added to work orders now has multiple users instead of one. You can pass a list of uuids in the workOrders.draft and workOrders.update endpoints and will receive a list of user entities in workOrders.info.
```

JordyDeGroeve: PP/B: `uuids`, no apostrophe needed

JordyDeGroeve (approved):

JordyDeGroeve (approved):

apiary.apib
```diff
@@ -3877,7 +3882,7 @@ Add a draft work order.
        + distance (Distance, optional)
        + time (array, optional)
            + (object)
-               + user_id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string, required)
+               + users: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a`,`f8a57a6f-dd1e-41a3-b8d3-428663f1d09e` (array[string], required)
```
mrtus: Q: Should we call it `user_ids`?

jenssegers: While I think `users` looks nicer, `user_ids` is the way to go for consistency.

WouterSioen: I couldn't find an example of some place doing the same. We have `participants` f.e. in projects, but in there, you get a VO with a participant and a role, which is not exactly the same as here.

Not sure if we should also make the structure like this to make it more extendible in the future if we would add metadata to each user?

```
{
    "users": [
        {
             "type": "user",
             "id": "4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a"
        },
        {
             "type": "user",
             "id": "f8a57a6f-dd1e-41a3-b8d3-428663"
        }
    ]
}
```

This way we're sure we'll never have to bc break this anymore in the future?

jenssegers: If we want to add additional data to the user then you will need a similar structure as the participants for projects. I remember us saying that we should support both ways eg. passing either a `user_id` or a `user` reference with type/id. But we ended up only implementing the short form. The short form was easier for users that wanted to requests using "regular" non-json GET/POST requests.

Also, see https://github.com/teamleadercrm/api/blob/master/GUIDELINES.md#creating-relationships.

But I feel like the mix of usages between the short `_id` form and the full `id/type` reference is only causing confusion now.

WouterSioen: Yep, I have the same feeling too.  We could make it like this, but not sure if that's better than just following the "read structure" everywhere.

### write endpoints:

```json
{
    "users": [
        {
             "user_id": "4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a"
        },
        {
             "user_id": "f8a57a6f-dd1e-41a3-b8d3-428663"
        }
    ]
}
```

### read endpoints:

```json
{
    "users": [
        {
             "type": "user",
             "id": "4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a"
        },
        {
             "type": "user",
             "id": "f8a57a6f-dd1e-41a3-b8d3-428663"
        }
    ]
}
```

jenssegers: If we want to go for the short form I think we should do:

```
{
    "users_ids": [
        "4ec55a8c-2d80-472a-a9c2-5ff5ee",
        "f8a57a6f-dd1e-41a3-b8d3-428663"
    ]
}
```

In that way it's similar to the `ids` filter we have.

But my ❤️  says to go for the second one and have consistency between write and read endpoints. But that's a big general API change.

WouterSioen: @jenssegers I think it would be nice to keep it in json objects to allow extensibility. I wouldn't be surprised if work order assignees could get roles in the future too.

WouterSioen: Let's poke @stivni here to help make this decision.

apiary.apib
```diff
@@ -429,7 +429,7 @@ We list all backwards-incompatible changes here. As described above, new additio
 
 #### 2019-07-03
 
-- time added to work orders now has multiple users instead of one. You can pass a list of uuid's in the workOrders.draft and workOrders.update endpoints and will receive a list of user entities in workOrders.info.
+- Time added to work orders now has multiple users instead of one. You can pass a list of uuids in the workOrders.draft and workOrders.update endpoints and will receive a list of user entities in workOrders.info.
```
dietrichm: B: "uuids" should be "UUIDs" in my opinion. Also the source file was not updated here.

apiary.apib
dietrichm: From previous experience (😅) we should keep the short form right now and switch (covered within a BC layer) to the other format when we need it. But this is something that can be discussed with Stijn in the TA day after tomorrow.

dietrichm (changes requested):
https://github.com/teamleadercrm/api/pull/425/files#r290272460

apiary.apib
jenssegers: @WouterSioen Even with the object format, you will need to change the structure if you decide to add more data. It will become something like this then:

```
{
    "users": [
        {
            "user": {
                "type": "user",
                "id": "4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a"
            },
            "role": "something"
        },
        {
            "user": {
                "type": "user",
                "id": "f8a57a6f-dd1e-41a3-b8d3-428663"
            }
            "role": "something"
        }
    ]
}
```

dietrichm (approved):

JordyDeGroeve (approved):

apiary.apib
stivni: Actually not strongly opinionated on this one, because we're loosing anyways :)
But if you would ask, I would go for the simple format for now.

WouterSioen:
Updated the proposal to use the most simple solution, and be consistent with other endpoints (so _id in writes and full objects in reads).

dietrichm (approved):

dietrichm:
✅ Checked

jenssegers (approved):

apiary.apib
WouterSioen: going for `users` in the read endpoint and `user_ids` in the write endpoints for now for consistency. We can always adapt later.

tiagobastos (approved):

PR #426, opened 2019-06-04
Add budget information at project level

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-1979

### What has been done
- Added budget information at project level
- Documented the changes in the changelog

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,18 +371,14 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
-- We added `budget` to `projects.list` and `projects.info`.
+- In `projects.list` and `projects.info`, we added the `budget` property that contains `provided`, `spent` and `remaining` budget.
+  This property is only available for users that have access to budgets on projects.
 
 - We added `sent` as a boolean to `invoices.info` and `invoices.list`
 
 - We added an optional filter on `ids` to `invoices.list`.
 
 - We added `creditNote.booked` and `product.added` to the available webhooks list.
-
-#### June 2019
-
-- In `projects.list` and `projects.info`, we added the `budget` property that contains `provided`, `spent` and `remaining` budget.
-  This property is only available for users that have access to Budgets on Projects.
 
 #### April 2019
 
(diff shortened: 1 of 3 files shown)


joaoinacio (dismissed):

ArnaudWeyts (dismissed):

adognameddog (dismissed):

apiary.apib
```diff
@@ -371,12 +371,19 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added `budget` to `projects.list` and `projects.info`.
```
stivni: You don't need to document it twice... adding it to the latest section is fine.
Every month we will group "latest" stuff to the specific month, so those entries stay stable and people don't have to reread them all the time..

stivni (dismissed):
The only remark I have, is that the provided budget is in theory the only updateable field.
So if we want to add that to an API, we should think about how we do that, so that that is clear...
(Is this planned, for sake of completeness of the API?)

jenssegers (dismissed):

joaoinacio (approved):

ArnaudWeyts (approved):

ketlane:
> The only remark I have, is that the provided budget is in theory the only updateable field.
> So if we want to add that to an API, we should think about how we do that, so that that is clear...
> (Is this planned, for sake of completeness of the API?)

The old API allows to specify a key called `project_budget` when creating a project. The new API does not currently allow to specify the budget. I propose to use a key called `budget`. If we go with that, I think it makes sense that `projects.info` returns a key called `budget.provided` (as it does now).

However, we haven't planned to allow specifying a budget when adding a project in API v2. I guess we could include that in the near future (with a way to also update this value when updating the project), it's not much work and should close this loop.

jenssegers (approved):

adognameddog (approved):

ketlane:
Closed in favour of https://github.com/teamleadercrm/api/pull/454

PR #428, opened 2019-06-04
[CAL-1477] Added due_by and due_from filters to tasks.list

r1dotl:
Added due_by and due_from filters to tasks.list

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -11,8 +11,8 @@ Get a list of tasks.
             + ids: `9c475de2-1e99-0328-bc19-271b27e921d8`,`6325b4e5-c6a6-060b-b010-b81892e82fa7` (array[string], optional)
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
             + completed: true (boolean, optional)
-            + due_after: `2019-02-04` (string, optional)
-            + due_before: `2019-10-20` (string, optional)
+            + due_after: `2019-02-04` (string, optional, nullable)
+            + due_before: `2019-10-20` (string, optional, nullable)
         + page (Page, optional)
         + sort (array, optional)
             + (object)
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -11,6 +11,8 @@ Get a list of tasks.
             + ids: `9c475de2-1e99-0328-bc19-271b27e921d8`,`6325b4e5-c6a6-060b-b010-b81892e82fa7` (array[string], optional)
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
             + completed: true (boolean, optional)
+            + due_after: `2019-02-04T16:44:33+00:00` (string, optional)
+            + due_before: `2019-10-20T00:00:00+00:00` (string, optional)
         + page (Page, optional)
         + sort (array, optional)
             + (object)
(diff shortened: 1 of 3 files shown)


stivni (dismissed):
👌 except for the whitespace changes you made 🙈

jenssegers (dismissed):

WouterSioen (dismissed):

src/08-tasks/tasks.apib
```diff
@@ -11,6 +11,8 @@ Get a list of tasks.
             + ids: `9c475de2-1e99-0328-bc19-271b27e921d8`,`6325b4e5-c6a6-060b-b010-b81892e82fa7` (array[string], optional)
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
             + completed: true (boolean, optional)
+            + due_after: `2019-02-04` (string, optional)
+            + due_before: `2019-10-20` (string, optional)
```
psrpinto: We got some [feedback](https://github.com/teamleadercrm/core/pull/10241#discussion_r290587243) on the implementation PR about naming these fields `due_since` and `due_until`, since they are inclusive. I tend to agree but doubting about the naming since I don't see any other instances of `_until`.

`due_since` and `due_until` make sense to me, just wondering what others think @stivni @jenssegers

src/08-tasks/tasks.apib
```diff
@@ -11,8 +11,8 @@ Get a list of tasks.
             + ids: `9c475de2-1e99-0328-bc19-271b27e921d8`,`6325b4e5-c6a6-060b-b010-b81892e82fa7` (array[string], optional)
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Returns Tasks that are assigned to this user or to a team to which this user belongs.
             + completed: true (boolean, optional)
-            + due_after: `2019-02-04` (string, optional)
-            + due_before: `2019-10-20` (string, optional)
+            + due_after: `2019-02-04` (string, optional, nullable)
+            + due_before: `2019-10-20` (string, optional, nullable)
```
psrpinto: B: I don't think these filters should be `nullable`, for me that would mean "give me tasks with no due date" but due date is a required field of a Task.

src/08-tasks/tasks.apib
stivni: Personally, I think `since` implies "past". It's hard to say it's "due since 1 January 2020"... what does that mean? 🤔 

Maybe it is fine to not make it inclusive? Or maybe call them `due_on_or_after` etc?

jenssegers (dismissed):

WouterSioen (dismissed):
Not entirely agreeing with the naming since they are inclusive and seem exclusive to me, but agreeing with the proposal in general

mikeverf (dismissed):

psrpinto:
`due_by` and `due_from` won the popular vote, we'll update this PR accordingly.

jenssegers:
@regularjack That would break consistency with the events and timetracking API. Although those are indeed times and not dates.

psrpinto:
These are indeed dates, not datetimes. What should we do @jenssegers, do we use `due_by` and `due_from`? If not, how can we reach an agreement about this?

psrpinto:
@jenssegers @stivni can we get this approved? We need to get it out the door this sprint.

mikeverf (dismissed):

WouterSioen (approved):

mikeverf (approved):

psrpinto:
Didn't hear back from @jenssegers so committing to `due_by` and `due_from`.

jenssegers:
My initial though was to keep consistency with the other date time related filters to make it simple. I'm not sure if renaming it will make it more obvious if the start/end is inclusive or not. To me the rename didn't solve that.

But if anyone agrees that this is a better fit in this context I'm ok with it.

stivni:
For me, the naming makes it more clear. So 🚢it

jenssegers (approved):

PR #430, opened 2019-06-06
Fix rendering issue with estimated_duration

psrpinto:
Before:
Screenshot 2019-06-06 at 10 26 24

After:
Screenshot 2019-06-06 at 10 27 01

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3314,7 +3314,7 @@ Get a list of tasks.
                 + completed_at: `2016-02-04T16:44:33+00:00` (string)
                 + due_at: `2016-02-04T16:44:33+00:00` (string)
                 + estimated_duration (object)
-                    + unit: `min` (enum[string])
+                    + unit (enum[string])
                         + Members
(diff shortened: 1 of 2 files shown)


mikeverf (approved):

jenssegers (approved):

jenssegers:
@regularjack Could we introduce a change here to make the duration consistent with the other endpoints and return the duration in seconds? (out of scope of this PR of course, just a question)

r1dotl (approved):

JorenSaey (approved):

psrpinto:
> Could we introduce a change here to make the duration consistent with the other endpoints and return the duration in seconds?

@jenssegers It was a deliberate choice to use minutes, this was done in https://github.com/teamleadercrm/api/pull/309:

> Even though we want to standardise on seconds everywhere, it just doesn't make sense here since in the Web UI we're using minutes. Using seconds on the API would lead to unexpected behaviour since we'd have to round to minutes.

PR #432, opened 2019-06-06
[CAL-1492] Add task_id filter to events.list.

r1dotl:
Implementation: https://github.com/teamleadercrm/core/pull/10300

`task_id` filter was added to `/events.list`

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,6 +371,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- `task_id` filter has been added to `/events.list`.
+
 - In order to link milestones and deals to a task, we added `milestone_id` and `deal_id` as optional parameters to `tasks.create` and `tasks.update`.
 
 - `tasks.list` and `tasks.info` now return the milestone and deal objects.
@@ -1961,6 +1963,7 @@ Get a list of calendar events.
                         + contact
                         + company
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + task_id: `5309c8c9-e313-47b0-90ba-6850c3dc3e33` (string, optional)
         + page (Page, optional)
         + sort (array, optional)
             + (object)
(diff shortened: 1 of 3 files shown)


WouterSioen (dismissed):

mikeverf (dismissed):

jenssegers:
Can we use `todo_id` here?

jenssegers:
The event object is returning a `todo`. So either we need to go for `todo_id` here or rename the `todo` to `task` in the response imo.

r1dotl:
> The event object is returning a `todo`. So either we need to go for `todo_id` here or rename the `todo` to `task` in the response imo.

We are working on renaming everything to task instead of todo.

JorenSaey (dismissed):

psrpinto:
> The event object is returning a todo. So either we need to go for todo_id here or rename the todo to task in the response imo.

@jenssegers That's indeed our plan, we'll be renaming `todo` in the Event response to `task`, in the next API version.

psrpinto:
conflicts 🚑

jenssegers:
@regularjack Will this be released together with that rename? Because now it's going to become even more confusing having both task and todo in 1 endpoint.

psrpinto:
@jenssegers The goal is to release the rename in the next version, 2019-07-03. We weren't planning on holding this PR until that version, but can do so if you think it's best. In that case, this PR would change to target branch `2019-07-03` instead of master.

stivni:
Another solution might be to name the filter `todo` right now and change it together with the other BC break?

psrpinto:
@stivni That would imply a BC-compatibility layer, so I'd prefer the simpler option of releasing this in version 2019-07-03.

@r1dotl what do you think? If you agree, we'd change the target branch of this to `2019-07-03`.

r1dotl:
> @stivni That would imply a BC-compatibility layer, so I'd prefer the simpler option of releasing this in version 2019-07-03.
> 
> @r1dotl what do you think? If you agree, we'd change the target branch of this to `2019-07-03`.

I agree with releasing it with the 2019-07-03 version.

r1dotl:
Closing this one and will open one based on 2019-07-03 branch.

PR #439, opened 2019-06-19
[CAL-1554] Add cancel_events attribute to tasks.delete

Jdww:
Add `cancel_events` option which would also cancel the events linked to the task

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3472,7 +3472,7 @@ Delete a task.
 
     + Attributes (object)
         + id: `5f0afd8a-8a40-48a4-bbe6-7d0e9c61bb6d` (string, required)
-        + deleteEvents: `false` (boolean, optional)
+        + cancel_events: `false` (boolean, optional)
 
 + Response 204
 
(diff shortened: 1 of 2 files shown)


r1dotl:
@jenssegers, @stivni have we ever done something like this before?

jenssegers:
@r1dotl Not really. My first guess would be that we should use the recycle bin for all deleted things. Why do you need this?

psrpinto:
@jenssegers We need this because deleting a Task should cancel all "schedulings" of it too. 

"scheduling" = an Event that is linked to that Task.

apiary.apib
```diff
@@ -3472,6 +3472,7 @@ Delete a task.
 
     + Attributes (object)
         + id: `5f0afd8a-8a40-48a4-bbe6-7d0e9c61bb6d` (string, required)
+        + deleteEvents: `false` (boolean, optional)
```
mrtus: We tend to use snake cased values.

jenssegers:
At first sight this seems like putting a lot of responsibility into the hands of the one who calls this endpoint. What is the expected behaviour if you delete a task? Should we set the default behaviour to not delete events linked to it and solve that through the UI?

Or do customers use this a lot? Do we have any tracking on this?

Also poke @stivni

psrpinto:
Closing since we decided to take a different approach.

PR #440, opened 2019-06-24
Add time to the workOrders.list endpoint

WouterSioen:
We'll want to display time in work orders overviews (f.e. in the mobile
app, but might be useful for other integrations too). Let's thus include
it in the workOrders.list endpoint.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3788,7 +3788,7 @@ Get a list of work orders.
                 + location (Address, nullable)
                 + distance (Distance, nullable)
                 + signed: false (boolean)
-                + time (array)
+                + time (array) - Only included with request parameter `includes=time`.
                     + (object)
                         + users (array)
                             + (object)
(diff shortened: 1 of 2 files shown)


JordyDeGroeve (approved):

JordyDeGroeve (approved):

stivni (approved):

dietrichm (approved):

apiary.apib
```diff
@@ -3788,6 +3788,15 @@ Get a list of work orders.
                 + location (Address, nullable)
                 + distance (Distance, nullable)
                 + signed: false (boolean)
+                + time (array) - Only included with request parameter `includes=time`.
+                    + (object)
+                        + users (array)
+                            + (object)
+                                + type: `user` (string)
+                                + id: `4ec55a8c-2d80-472a-a9c2-5ff5ee7eac6a` (string)
+                        + started_at: `2019-02-27T10:01:49+00:00` (string)
```
jenssegers, line 3797: Nice to have; also add `ended_at` like we have for time tracking.

tiagobastos (approved):

jenssegers (approved):

apiary.apib
WouterSioen: Seems like a separate feature, I’ll take it up with product

PR #442, opened 2019-06-24
[CAL-1546][CAL-1547] tasks.complete and tasks.reopen

psrpinto:


apiary.apib
```diff
@@ -3464,6 +3466,28 @@ Update a task.
...
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `46156648-87c6-478d-8aa7-1dc3a00dacab` (string, required)
+
++ Response 204
+
+### tasks.reopen[POST /tasks.reopen]
+
+Reopen a Task that had been marked as complete.
```
stivni: PP

```suggestion
Reopen a task that had been marked as complete. 
```

stivni (dismissed):

mikeverf (dismissed):

JorenSaey (approved):

apiary.apib
```diff
@@ -3464,6 +3466,28 @@ Update a task.
 
 + Response 204
 
+### tasks.complete [POST /tasks.complete]
+
+Mark a task as complete.
```
jenssegers, line 3471:
```suggestion
Mark a task as completed.
```

jenssegers: What is the proper spelling in this case actually? I kinda see both when I google for this.

jenssegers (approved):

apiary.apib
psrpinto: I think they can both be used.

Actually, thinking a bit more about this, we pretty much always use the term *done* instead of complete. Would it make sense to call the endpoint `/tasks.done`? Or I guess in that case it would need to be `/tasks.markAsDone`? 

I used `/tasks.complete` because that's what we have in api-future, but we don't really use that term anywhere.

psrpinto: ^ I guess if we would do this we would also need to rename the property on the task from `complete` to `done`.

r1dotl (approved):

PR #445, opened 2019-07-08
Filter milestones.list on term

ArnaudWeyts:
Only searches on the title of the milestone

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3181,6 +3181,7 @@ Get a list of project milestones.
                     + closed
             + due_before: `2017-01-01` (string, optional)
             + due_after: `2017-01-01` (string, optional)
+            + term: `coffee` (string, optional) - Searches for a term in title
         + page (Page, optional)
         + sort (array, optional)
             + (object)
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,6 +371,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added a `term` filter to `milestone.list`.
+
 - We added `tasks.complete` and `tasks.reopen` endpoints.
 
 - We added a `tasks.schedule` endpoint.
--- a/src/07-projects/milestones.apib
+++ b/src/07-projects/milestones.apib
@@ -20,6 +20,7 @@ Get a list of project milestones.
                     + closed
             + due_before: `2017-01-01` (string, optional)
             + due_after: `2017-01-01` (string, optional)
+            + term: `coffee` (string, optional) - Searches for a term in title
         + page (Page, optional)
         + sort (array, optional)
             + (object)
--- a/src/changes-backwards-compatible.apib
+++ b/src/changes-backwards-compatible.apib
@@ -6,6 +6,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added a `term` filter to `milestone.list`.
+
 - We added `tasks.complete` and `tasks.reopen` endpoints.
 
 - We added a `tasks.schedule` endpoint.


joaoinacio (dismissed):

mikeverf (dismissed):

joaoinacio (dismissed):

adognameddog (dismissed):

apiary.apib
```diff
@@ -3181,6 +3183,7 @@ Get a list of project milestones.
                     + closed
             + due_before: `2017-01-01` (string, optional)
             + due_after: `2017-01-01` (string, optional)
+            + term: `coffee` (string, optional) - Searches for a term in title
```
jenssegers, line 3204: If it's only searching on the title let's call the filter `title`. The `term` filter on contacts/companies searches on more than one property eg. name, phone number, address, ...

```suggestion
            + title: `coffee` (string, optional) - Searches for a term in title
```

jenssegers (changes requested):
Rename `term` filter to `title`.

apiary.apib
ketlane: Hm, but if we decide to also search in the description (or other fields, like the project title/description) in one month, we will need to rename this to 'term', then, causing a breaking change/API version change.

Besides that, as an API consumer, I find it weird/inconsistent to have some endpoints use `term` as keyword search field and other endpoints in the same API use another field for keyword search. The slight difference that in one endpoint the search is done on two fields while in another endpoint the search is done in only one field (for now) is not enough to sustain such an inconsistency, in my point of view.

adognameddog:
> Rename `term` filter to `title`.

I agree and disagree at the same time. Title is more clear as to what you are actually searching in, but `term` wins it for me. To me, if you name that filter `title`, it feels like you have to provide the full and correct title to filter on it, wether as `term` indicates I can filter on part of the title.

Moreover, if we include more fields to search in, like description, we have breaking change, like Catalin mentioned.

So actually I disagree 🙂

apiary.apib
jenssegers: Why not have both? Add a `title` filter now to only search on the title. And if we want a more general search term add `term`. No breaking changes needed.

The more explicit the filters are the better. If you want to search by title, use a `title` filter. If you want to search in all properties, then use `term`.

jenssegers: You can add a `term` filter if you want. Just make sure you keep in mind that integrators will use it to search on more than the title.

jenssegers (dismissed):

jenssegers:
My PP would be to have both a `titel` and `term` filter so that the user of our API can choose how he wants to search.

adognameddog:
> My PP would be to have both a `titel` and `term` filter so that the user of our API can choose how he wants to search.

That's a good point. The request came from our esteemed app developers, so there is no direct need for it. I will ask them though, maybe they hadn't thought about searching on exact title.

joaoinacio:
Just an FYI, it seems that the requests to milestones.list (coming from the new TImeTracker UI, for example) have already been "using" the `term` filter for a while, before it was even implemented.

ketlane (approved):

mikeverf (approved):

jenssegers (approved):

adognameddog (approved):

adognameddog:
Manually checked ✅

joaoinacio (approved):

apiary.apib
stivni: Imo this depends on the current knowledge about which filters we're still going to add...
If searching on more than "title" is currently planned, it's OK to go for "term".
If nothing is planned, by yagni and call it "title".

stivni (approved):

PR #446, opened 2019-07-09
PRO-2092: add invoicing_method, amount to milestones.create/update

joaoinacio:


ketlane (dismissed):

ketlane:
Should we allow this to be specified also when adding a project?

apiary.apib
```diff
@@ -3262,6 +3265,12 @@ Create a new milestone.
         + due_on: `2018-01-01` (string, required)
         + name: `Initial setup` (string, required)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
+        + `invoicing_method`: `time_and_materials` (enum[string])
+            + Members
+                + time_and_materials
+                + fixed_price
+                + non_invoiceable
+        + `invoicing_amount`: `599.50` (number, optional) - Required when method is `fixed_price`.
```
adognameddog: Q: what happens when you provide an `invoicing_amount`, and the `invoicing_method` is *not* `fixed_price`? Maybe we should add "If the method is anything else, this value is ignored" or something?

ArnaudWeyts (dismissed):

apiary.apib
joaoinacio: > Q: what happens when you provide an `invoicing_amount`, and the `invoicing_method` is _not_ `fixed_price`? Maybe we should add "If the method is anything else, this value is ignored" or something?

Correct, the amount is ignored if method is not fixed_price. will update

ArnaudWeyts (approved):

src/07-projects/milestones.apib
```diff
@@ -101,6 +101,12 @@ Create a new milestone.
         + due_on: `2018-01-01` (string, required)
         + name: `Initial setup` (string, required)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
+        + `invoicing_method`: `time_and_materials` (enum[string])
```
jenssegers, line 104: Q: Could we name this something like `pricing_model`? We might have some mixed concepts between an invoicing method for invoicing endpoints which could be PayPal, credit card, bank transfer (I don't know) and invoicing method in this context. If this is indeed the correct term ignore the comment. But if we could avoid potential confusion it might be worth double checking :)

If I search for time-and-materials vs. fixed price they often refer to it as the project pricing model.

joaoinacio: > Q: Could we name this something like `pricing_model`? We might have some mixed concepts between an invoicing method for invoicing endpoints which could be PayPal, credit card, bank transfer (I don't know) and invoicing method in this context. If this is indeed the correct term ignore the comment. But if we could avoid potential confusion it might be worth double checking :)
> 
> If I search for time-and-materials vs. fixed price they often refer to it as the project pricing model.

My 2 cents:
We already provide this information on `milestones.list` and `milestones.info` endpoint, so it wouldn't make sense to have different (field) names for the same purpose/fields. 
The term that is used throughout the UI in project milestones is actually `billing method`.

jenssegers: I do prefer billing method over invoicing method indeed.

apiary.apib
```diff
@@ -3262,6 +3265,12 @@ Create a new milestone.
         + due_on: `2018-01-01` (string, required)
         + name: `Initial setup` (string, required)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
+        + `invoicing_method`: `time_and_materials` (enum[string])
```
ketlane, line 3268: Q: I noticed the other fields (i.e. responsible_user_id, name etc) are not encapsulated in backticks; why is this one?

ketlane (approved):

adognameddog:
We are taking a different approach to budgets and billing methods on milestones.

PR #449, opened 2019-07-09
Allow filtering timeTracking by subject

psrpinto:
Add a filter by `subject` to `timeTracking.list`.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3538,6 +3538,16 @@ Get a list of tracked time.
             + started_before: `2017-04-26T10:01:49+00:00` (string, optional)
             + ended_after: `2017-04-26T10:01:49+00:00` (string, optional)
             + ended_before: `2017-05-26T10:01:49+00:00` (string, optional)
+            + subject (object, optional)
+                + type (enum, required)
+                    + Members
+                        + company
+                        + contact
+                        + event
+                        + todo
+                        + milestone
+                        + ticket
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
         + page (Page, optional)
         + sort (array, optional)
             + (object)
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,6 +371,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added a filter by `subject` to `timeTracking.list`.
+
 - We added `tasks.complete` and `tasks.reopen` endpoints.
 
 - We added a `tasks.schedule` endpoint.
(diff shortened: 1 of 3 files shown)


tiagobastos (dismissed):

ketlane (dismissed):
Approved, but aren't we missing the release notes?

psrpinto:
Oops, forgot about the release notes. Fixed

jenssegers (dismissed):

joaoinacio (dismissed):

tiagobastos (dismissed):

ketlane (dismissed):

tiagobastos:
Merged master to resolve conflicts. 
Time-Tracking service was deployed with the filter on subject. 
So this can be merge after reviews.

apiary.apib
```diff
@@ -3571,6 +3573,16 @@ Get a list of tracked time.
             + started_before: `2017-04-26T10:01:49+00:00` (string, optional)
             + ended_after: `2017-04-26T10:01:49+00:00` (string, optional)
             + ended_before: `2017-05-26T10:01:49+00:00` (string, optional)
+            + subject (object, optional)
+                + type (enum, required)
+                    + Members
+                        + company
+                        + contact
+                        + event
+                        + todo
```
joaoinacio, line 3582: Q: shouldn't this be `task` ? I thought we were getting/trying to get rid of `todo` (and allow it just for compatibility).

See also https://github.com/teamleadercrm/api-gateway/commit/75bae24a7e8aad2fe612549e53722e8630e3ecfb

PP: document  `task`, maybe allow `todo` for BC only.

tiagobastos: @r1dotl 👆

r1dotl: The thing is, on the database it is stored as `todo`, so in order to use `task` here instead of `todo` we would have to convert it somewhere in the code, from `task` to `todo`, because the `subject_type` (db column on crm_timetracking) only allows `todo` and not `task`.

r1dotl (approved):

apiary.apib
joaoinacio: > The thing is, on the database it is stored as `todo`, so in order to use `task` here instead of `todo` we would have to convert it somewhere in the code, from `task` to `todo`, because the `subject_type` (db column on crm_timetracking) only allows `todo` and not `task`.

Ok, but that's just an implementation detail, right? 
(AFAICT it should need an update in `TimeTrackingSearchFiltersFactory` or `SubjectType` VO,  as well as the request validator, plus some more tests).

r1dotl: Yep thats true. Let me discuss it with @marcwestermann and @Jdww

JordyDeGroeve: Any update on this?

JordyDeGroeve (approved):

apiary.apib
jenssegers: My guess is that this should become `task`, since we're going to have `tasks.list` and `events.list` endpoints. Right?

r1dotl: @joaoinacio @JordyDeGroeve @jenssegers yup, we have a ticket for the next sprint dedicated to this subject :)

tiagobastos (approved):

PR #450, opened 2019-07-10
Add term filter to tasks.list

psrpinto:
This has been implemented but the documentation was missing.


apiary.apib
```diff
@@ -3301,6 +3303,7 @@ Get a list of tasks.
             + completed: true (boolean, optional)
             + due_by: `2019-02-04` (string, optional)
             + due_from: `2019-10-20` (string, optional)
+            + term: `website design` (string, optional) - Searches for a term in the description.
```
jenssegers, line 3336: If it only searches in the description, shouldn't we name the filter `description`?

jenssegers (dismissed):

apiary.apib
psrpinto: It currently searches on description but the idea is that it can be extended in the future. Should I remove the `Searches for a term in the description` bit?

mikeverf (dismissed):

r1dotl (dismissed):

r1dotl (approved):

apiary.apib
jenssegers: Ideally, we would have both a `description` and `term` filter so that the end-user can choose how he wants to filter

mikeverf (approved):

jenssegers (approved):

apiary.apib
psrpinto: Will make sure to consider that once we extend the search to other fields.

PR #453, opened 2019-07-22
Budget on `projects.info` [PRO-2147]

adognameddog:
We want to expose our Better Budgets™ values to the api. This will only be available for users with access to project budgets. The spent budget is divided amongst time and materials. That division does not make sense for the other values.


adognameddog:
Hmmm, this should only be one commit :-/ Lemme try this again

PR #458, opened 2019-07-29
Add time_tracked to projects.info [PRO-2197]

adognameddog:
We added `tracked_time` to `projects.info`, with a breakdown of billable and non-billable.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,6 +371,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added `tracked_time` to `projects.info`, with a breakdown of billable and non-billable hours.
+
 - We added a filter by `subject` to `timeTracking.list`.
 
 - A `task_id` filter has been added to `/events.list`.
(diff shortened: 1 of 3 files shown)


joaoinacio (dismissed):

apiary.apib
```diff
@@ -3080,6 +3082,22 @@ Get details for a single project.
...
+                    + unit (enum)
+                        + Members
+                            + h - Hours
+                    + value: `10.25` (number)
+                + billable (object)
+                    + unit (enum)
+                        + Members
+                            + h - Hours
+                    + value: `7.50` (number)
+                + `non-billable` (object)
```
ArnaudWeyts: B: this should be `non_billable`

apiary.apib
```diff
@@ -3080,6 +3082,22 @@ Get details for a single project.
...
+                + total (object)
+                    + unit (enum)
+                        + Members
+                            + h - Hours
+                    + value: `10.25` (number)
+                + billable (object)
+                    + unit (enum)
+                        + Members
+                            + h - Hours
+                    + value: `7.50` (number)
```
ArnaudWeyts: ~~Should this be renamed to `quantity` to keep it consistent with how we'll do it with `projectItems`? We also have a unit and quantity key there~~

Cleared up offline: this is how it's defined for tasks, so we should probably adjust it for `projectItems` instead

ArnaudWeyts (changes requested):
See comment

apiary.apib
adognameddog: No, this is consistent with other endpoints. If anything, we should change it in `projectItems` 🙂

ArnaudWeyts (dismissed):

ArnaudWeyts:
✅ Manually checked

ArnaudWeyts (approved):

ketlane (approved):

joaoinacio (approved):

PR #462, opened 2019-08-05
archive the changes for July

stivni:
Since we want to put the changes of July under a July header, lets add this.

(Note on why I don't make an August header straight away: because imo the latest section is mutable and the monthly sections are not. In that way people who come read now and then, know they don't have to look at _the past_ because they already read that... but are going to rer-read _Latest_)

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -371,6 +371,10 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+
+
+#### July 2019
+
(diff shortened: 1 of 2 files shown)


JordyDeGroeve (approved):

steftrenson (approved):

dietrichm (approved):

dietrichm:
✅ Checked

ketlane (approved):

PR #471, opened 2019-09-24
Billing method and budget in milestones.create [PRO-2436]

adognameddog:


ArnaudWeyts (dismissed):

ketlane (dismissed):

apiary.apib
```diff
@@ -3313,6 +3319,18 @@ Create a new milestone.
...
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
+        + One Of
+            + Properties
+                + `billing_method`: `time_and_materials` (enum) - If omitted, `time_and_materials` is used
+                    + Members
+                        + `non_invoiceable`
+                        + `time_and_materials`
+                + budget (Money) - If omitted, the budget will default to `0.00`
+            + Properties
+                + `billing_method`: `fixed_price` (enum, required)
```
ArnaudWeyts: ~Q: The billing method isn't required here right? It's just the price when this billing method is selected~

Apparently this is the right way to do it, don't mind me 🙈

ketlane (approved):

ArnaudWeyts (approved):

apiary.apib
```diff
@@ -3313,6 +3319,18 @@ Create a new milestone.
         + due_on: `2018-01-01` (string, required)
         + name: `Initial setup` (string, required)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
+        + One Of
+            + Properties
+                + `billing_method`: `time_and_materials` (enum) - If omitted, `time_and_materials` is used
+                    + Members
+                        + `non_invoiceable`
+                        + `time_and_materials`
+                + budget (Money) - If omitted, the budget will default to `0.00`
```
jenssegers: Is this because of legacy reasons? Could we find a way to keep the budget nullable and return `null` instead of `0.00`?

adognameddog: That's quite a big change I'm afraid. But it is most definitely on our list. So yeah, it's because of legacy

jenssegers: Pitty :(

jenssegers (approved):

stivni (commented):
PP: I think apib has a way to document the default options of an enum and maybe even the default value of a number

ArnaudWeyts (approved):

ArnaudWeyts:
✅ Manually checked

PR #473, opened 2019-10-02
Add follow_up_due_on filter to deals.list

luisaterrabrito:


apiary.apib
```diff
@@ -1592,6 +1592,7 @@ Get a list of deals.
             + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22` (string, optional)
             + updated_since: `2018-02-11T16:45:30+00:00` (string, optional)
             + created_before: `2018-02-11T16:45:30+00:00` (string, optional)
+            + follow_up_due_on: `2018-02-11` (string, optional)
```
stivni, line 1595: Q : should we go for `follow_up_due_before` (and NTH : `follow_up_due_after`)?

stivni: In tasks it is called `due_by` and `due_from`, however

PR #474, opened 2019-10-15
Rename creditNote to credit_note in migrating

luisaterrabrito:


jenssegers:
Why? All our API types are camel cased.

I think we need to fix the code to reflect the documentation in this case.

luisaterrabrito:
Closing this and will allow both in the code

PR #480, opened 2019-10-30
milestones.delete [PRO-2747]

adognameddog:
What it says on the tin

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added `milestones.delete`.
+
 - We added a `estimated_closing_date` filter to `deals.list`.
 
 - We added a filter `created_before` to `deal.list`.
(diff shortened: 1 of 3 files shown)


ketlane (approved):

joaoinacio (approved):

ArnaudWeyts (approved):

apiary.apib
```diff
@@ -3360,6 +3362,17 @@ Update a milestone.
 
 + Response 204 (application/json)
 
+### milestones.delete [POST /milestones.delete]
+
+Delete a milestone.
```
jenssegers, line 3367: You are not able to delete the last milestone right? Can we add some more information here to help out the developers using the API? :)

jenssegers (approved):

stivni (approved):

apiary.apib
adognameddog: There are a lot of conditions that will prevent you from deleting a milestone. The error message is perfectly verbose and clear, so I thought to omit all the conditions

jenssegers: The earlier we can make it clear in the documentation the better though ;)

PR #482, opened 2019-12-11
[PRO-2919] Documentation: the tasks can now be unassigned

ketlane:
### Ticket
Link to ticket: https://teamleader.atlassian.net/browse/PRO-2919

We are now allowing tasks to be unassigned, behind a feature flag. Before switching the feature flag on for everyone, we're preparing the API documentation for this.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- The task assignee is now nullable.
+
 - We added `milestones.delete`.
 
 - We added a `estimated_closing_date` filter to `deals.list`.
@@ -524,10 +526,6 @@ _This will only affect the version for those API calls and won't affect any othe
 ### Changelog
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
-
-#### 2019-12-11
-
-- The task assignee is now nullable.
 
 #### 2019-10-09
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -525,6 +525,10 @@ _This will only affect the version for those API calls and won't affect any othe
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
 
+#### 2019-12-11
+
+- The task assignee is now nullable.
+
 #### 2019-10-09
 
 - We added `billing_method` and `budget`, and `billing_method` and `price` to `milestones.create`.
@@ -3521,7 +3525,7 @@ Create a new task.
                 + Members
                     + min - Minutes
             + value: `60` (number, required)
-        + assignee (object, optional)
+        + assignee (object, optional, nullable)
             + type: `user` (enum)
                 + Members
                     + user
@@ -3560,7 +3564,7 @@ Update a task.
                 + Members
                     + min - Minutes
             + value: `60` (number, required)
-        + assignee (object, optional)
+        + assignee (object, optional, nullable)
             + type: `user` (enum)
                 + Members
                     + user
--- a/apiary.apib
+++ b/apiary.apib
@@ -3433,7 +3433,7 @@ Get a list of tasks.
                 + work_type (object, nullable)
                     + type: `workType` (string)
                     + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
-                + assignee (object)
+                + assignee (object, nullable)
                     + type: `user` (enum)
                         + Members
                             + user
@@ -3481,7 +3481,7 @@ Get information about a task.
             + work_type (object, nullable)
                 + type: `workType` (string)
                 + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
-            + assignee (object)
+            + assignee (object, nullable)
                 + type: `user` (enum)
                     + Members
                         + user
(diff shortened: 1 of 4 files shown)


ArnaudWeyts (dismissed):

apiary.apib
```diff
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- The task assignee is now nullable.
```
joaoinacio: Q: Maybe it would make sense to add a small explanation? 

> The task assignee is now nullable (unassigned task).

joaoinacio (dismissed):

adognameddog (dismissed):

src/changes-backwards-incompatible.apib
```diff
@@ -3,6 +3,10 @@
 
 We list all backwards-incompatible changes here. As described above, new additions and forwards-compatible changes don’t need a new API version and can be found [here](#additions).
 
+#### 2019-12-11
+
+- The task assignee is now nullable.
```
psrpinto: Q: How will we maintain compatibility for consumers on earlier versions? If a task is created without assignee, a consumer on an earlier version calling `tasks.info` will get `null` as assignee, which is a BC break, right?

ketlane: After discussing this with @stivni we agreed to return the task assignee as `{ type: user, id: null }` for older versions, in case the task is unassigned. It's still not perfect, since previously you would always get a non-null id, but at least it respects the format. 

The reality is that, in this case, we didn't find a way to not break compatibility and allow unassigned tasks at the same time. We can only mitigate this by keeping the returned format for older versions.

apiary.apib
ketlane: How about the following:

> The task assignee is now nullable: you can create / update unassigned tasks by sending `null` as `assignee`. The `tasks.list` and `tasks.info` endpoints also return `null` as `assignee` in case the task is unassigned. 

> On API versions older than `2019-10-09`, the `assignee` is returned as  `{ type: user, id: null }` in case the task is unassigned.

src/changes-backwards-compatible.apib
```diff
@@ -6,7 +6,9 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
-- The task assignee is now nullable.
+- The task assignee is now nullable: you can create / update unassigned tasks by sending null as assignee. The tasks.list and tasks.info endpoints also return null as assignee in case the task is unassigned.
+
+- On API versions older than 2019-10-09, the assignee is returned as { type: user, id: null } in case the task is unassigned.
```
psrpinto, line 11: Q: It's on API versions up to and including 2019-10-09, right? (I'm assuming 2019-10-09 is the latest API version today).

ketlane: Not including. 2019-10-09 is the latest version. On 2019-10-09 and later it returns null; older than that, it returns type: user and id null.

psrpinto: I thought this was a breaking change, didn't realise this is the backwards-compatible changelog. Please ignore.

psrpinto (approved):

mikeverf (approved):

ArnaudWeyts (approved):

adognameddog (approved):

joaoinacio (approved):

tiagobastos (approved):

apiary.apib
```diff
@@ -374,6 +374,10 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- The task assignee is now nullable: you can create / update unassigned tasks by sending null as assignee. The tasks.list and tasks.info endpoints also return null as assignee in case the task is unassigned.
```
stivni, line 377: I imagine you can also not send the assignee?

src/changes-backwards-compatible.apib
stivni: But then the contract of the assignee field changes for `2019-10-09`? So it is breaking?

apiary.apib
ketlane: Yes, although that automagically assigns it to yourself, it seems :D. Not sure if this is a bug or it is intended: the assignee was previously (since the beginning of 2019) documented as optional (both in create and in update), so it's not something to add here.

ketlane: But the assignee was previously documented as optional (both in create and in update), so it's not something to add here.

apiary.apib
```diff
@@ -374,6 +374,10 @@ We list all backwards-compatible additions here. These are currently available i
...
+
+- On API versions older than 2019-10-09, the assignee is returned as { type: user, id: null } in case the task is unassigned.
```
jenssegers, line 379: To me `{ type: user, id: null }` looks a bit strange. I would think that sending back `null` as the id is as breaking as simply sending back `null` as the assignee.

Looks like it's very difficult to keep this working in a backward-compatible way.

Or we need to send back the creator as the assignee and tell our users to upgrade to a newer API version if they want to take advantage of the unassigned tasks.

If I remember correctly Stripe also forced you to upgrade to a newer version if you want to use newer features.

WDYT?

ArnaudWeyts: But sending the creator of the task as the assignee is literally wrong data. Also; wether you get `null` or a `UUID` doesn't really matter. What if the user has been deleted? Integrations should already be expecting UUID's to potentially not match a user. But they might not be expecting the reference object to be gone all together.

The only problem I see with this is our validation on `users.info` and `users.list` failing when you pass `null` instead of a valid `UUID`.

adognameddog:
Rebased to version branch (2020-01-01)

PR #483, opened 2020-01-06
IN-2383 Add filters estimated_closing_date range to deals.list

luisaterrabrito:


lowiebenoot (dismissed):

sanderbrugge (dismissed):

EOSullivanBerlin (dismissed):

JordyDeGroeve (dismissed):

src/03-deals/deals.apib
```diff
@@ -22,11 +22,13 @@ Get a list of deals.
...
             + estimated_closing_date: `2018-02-11` (string, optional, nullable)
+            + estimated_closing_date_before: `2020-01-31` (string, optional) - This date is inclusive
+            + estimated_closing_date_after: `2020-01-01` (string, optional) - This date is inclusive
             + One Of
                 + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22` (string, optional)
                 + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22`, `98b2863e-7b01-4232-82f5-ede1f0b9db24` (array[string], optional)
-            + updated_since: `2018-02-11T16:45:30+00:00` (string, optional)
-            + created_before: `2018-02-11T16:45:30+00:00` (string, optional)
+            + updated_since: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
+            + created_before: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
```
dietrichm, line 31: Q: is this how the endpoint filters at the moment? I'm not sure if we should make or keep this inclusive. "After" or "before" does specify that that date and time is not included, right?

src/03-deals/deals.apib
```diff
@@ -22,11 +22,13 @@ Get a list of deals.
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
             + phase_id: `1758e847-cc1f-04ce-bd17-5a852e8c01b4` (string, optional)
             + estimated_closing_date: `2018-02-11` (string, optional, nullable)
+            + estimated_closing_date_before: `2020-01-31` (string, optional) - This date is inclusive
+            + estimated_closing_date_after: `2020-01-01` (string, optional) - This date is inclusive
```
dietrichm: Q: same question here.

src/03-deals/deals.apib
dietrichm: It appears to be already implemented inclusively for these filters. This does behave differently than i.e. the filters for time tracking. 🤔

src/03-deals/deals.apib
luisaterrabrito: We checked other endpoints and discussed with Stijn

dietrichm: What is the reasoning to go for inclusive datetime properties?

To me this feels like adapting the API docs to the way it is currently implemented. This further introduces inconsistencies across the API plus the property name does not reflect the way it behaves (hence the need to add the comments to state it's inclusive).

dietrichm: B: As mentioned, let's go for the suffixes _from and _until to make it more explicit the dates are inclusive rather than exclusive. ~Then we can also remove the comment on the property.~ Let's keep that one for clarity.

~Strong PP: example dates should be datetimes (they are in the atom date time format like the other ones).~ As the values to be filtered are only dates without time, it makes sense that the filters do not accept times (for now).

dietrichm (changes requested):
Let's rename the properties and make sure the back-end is updated (as far as merged PRs concern).

dietrichm (approved):

sanderbrugge (approved):

EOSullivanBerlin (approved):

lowiebenoot (approved):

PR #484, opened 2020-01-13
Include related workOrder in event.info

JordyDeGroeve:


src/04-calendar/events.apib
```diff
@@ -109,6 +109,9 @@ Get details for a single calendar event.
             + starts_at: `2016-02-04T16:00:00+00:00` (string)
             + ends_at: `2016-02-04T18:00:00+00:00` (string)
             + location (string)
+            + related_work_order (object) - Only included with request parameter `includes=work_order`
```
JordyDeGroeve: Q: do we want to link only 1? Or should we return a list of linked workOrders?
(for now, it is only possible to link 1 workOrder from the UI. But I assume we want to enable multiple draft WOs for a meeting in the future.)

dietrichm: B: I'd rather say this fits in the existing `links` structure which holds a collection of interlinked entities.

A rather specific property for just (one, indeed) work order is pretty limiting and will get messy if we copy this pattern for more entity types.

dietrichm (dismissed):

tiagobastos:
> B: I'd rather say this fits in the existing links structure which holds a collection of interlinked entities.

@dietrichm isn't the `links` more tied to customers linked to that event? 
We looked into it but discarded it because of that.

JordyDeGroeve:
@dietrichm As mentioned by Tiago, the `links` part was our first option. But this requires us to save the **_workOrder <-> meeting_**-link in the link table.
And this does not feel like the correct thing to do

Can you clear up what the purpose of that table is exactly? Maybe we are wrong thinking that does not belong there

dietrichm:
> @dietrichm isn't the `links` more tied to customers linked to that event?
> We looked into it but discarded it because of that.

Right now it is, but it could be extended to be linked to more entities, right?

> @dietrichm As mentioned by Tiago, the `links` part was our first option. But this requires us to save the **_workOrder <-> meeting_**-link in the `cal_event_links` table.
> And this does not feel like the correct thing to do
> 
> Can you clear up what the purpose of that table is exactly? Maybe we are wrong thinking that does not belong there

Do you feel the API design should adapt itself to the way it's currently implemented in the back-end? I feel like this should not be the case. The API should be as clear and "logical" as possible regardless. And with that POV a calendar event could be linked to a work order.

If back-end implementation is making this hard to implement we should consider some refactorings, but we should not discuss that in a public repository.

tiagobastos (approved):

dietrichm (approved):

Kemosabert (approved):

kevinwaelkens (approved):

PR #492, opened 2020-01-31
Added a new return state for workOrder.download

nielsvermaut:


tiagobastos:
Q: I think we should just add it to `Status codes`. 🤔 

I'm checking and I can't find an API response that as multiple `+ Response ...` responses in it. Maybe we just return the successful one?

Kemosabert (approved):

kevinwaelkens (approved):

nielsvermaut:
> I'm checking and I can't find an API response that as multiple `+ Response ...` responses in it. Maybe we just return the successful one?

I feel that that would kinda render the docs flaky at best, if we'd only document the flows that are "most happy", instead of all supported flows. In my opinion, we should also document the invalid flows that are covered in our business logic, like contacts that are not found.

tiagobastos:
> > I'm checking and I can't find an API response that as multiple `+ Response ...` responses in it. Maybe we just return the successful one?
> 
> I feel that that would kinda render the docs flaky at best, if we'd only document the flows that are "most happy", instead of all supported flows. In my opinion, we should also document the invalid flows that are covered in our business logic, like contacts that are not found.

That makes sense indeed. But we don't do it right now. So shouldn't we strive for consistency and iterate on that after?

JordyDeGroeve (approved):

tiagobastos:
Also, we should compile the main apiary file

tiagobastos:
By removing the `work-orders` branch I closed this PR without realizing 🙈 
Sorry @nielsvermaut! Any follow up we should do on it?

PR #493, opened 2020-01-31
Adjust documentation for deals reference property

dietrichm:
The deal reference is a number, not a string value. Let's fix this in the documentation.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1645,7 +1645,7 @@ Get a list of deals.
             + (object)
                 + id: `6e7fe84d-d4b3-4723-abae-9bc082d8f65a` (string)
                 + title: `Interesting deal` (string)
-                + reference: `2017/2` (string)
+                + reference: `173` (number)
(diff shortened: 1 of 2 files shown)


sanderbrugge (approved):

luisaterrabrito (approved):

EOSullivanBerlin (approved):

steftrenson:
If I'm not mistaken we made it a string intentionally. Right now we don't allow external references, but in the future we might and I think we would allow strings as references then. @stivni what do you think?

dietrichm:
> If I'm not mistaken we made it a string intentionally. Right now we don't allow external references, but in the future we might and I think we would allow strings as references then. @stivni what do you think?

Good insights. It appears the value is always casted to a string indeed. Let's wait for @stivni's take here.

dietrichm:
Closing because it looks like this was intentional, so let's not shoot ourselves in our future feet (or something like that 😬).

PR #497, opened 2020-02-13
Expose the freeze time tracking preference on users.me

ketlane:
When the time tracking feature is enabled, there is a setting in the Teamleader UI that prevents users from being able to add or modify time tracking entries after a certain amount of days has passed since the time tracking's end date:

image

image

However, this preference is not made available to API users yet. 

This PR proposes to expose the preference on the `users.me` endpoint, so that UIs can disable editing time tracking entries when the editing cutoff time has passed.

More details on this functionality:
* the account administrators (super users) should still be able to modify entries
* when teams are used, the team leaders should still be able to modify entries for their team members, but not for themselves and not for any administrators in the team

The limitation will also be implemented on the time tracking add & update endpoints.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -732,8 +732,8 @@ Get the current authenticated user.
                 + historic_time_tracking_limit (object, nullable)
                     + unit (enum)
                         + Members
-                            + day - Days
-                    + value: 7 (number)
+                            + hour - Hours
+                    + value: 24 (number)
 
 ### users.list [GET /users.list]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -729,8 +729,11 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
-                + time_tracking_freezing: true (boolean)
-                + time_tracking_freezing_cutoff_days: 7 (number)
+                + historic_time_tracking_limit (object, nullable)
+                    + unit (enum)
+                        + Members
+                            + day - Days
+                    + value: 7 (number)
 
 ### users.list [GET /users.list]
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We exposed the "Freeze time tracking after a number of days" preference in `users.me`.
+
 - We added format `ubl/e-fff` to `invoices.download`.
 
 - We added filters `estimated_closing_date_from` and `estimated_closing_date_until` to `deals.list`.
@@ -727,6 +729,8 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
+                + time_tracking_freezing: true (boolean)
+                + time_tracking_freezing_cutoff_days: 7 (number)
 
 ### users.list [GET /users.list]
 
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -40,6 +40,8 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
+                + time_tracking_freezing: true (boolean)
+                + time_tracking_freezing_cutoff_days: 7 (integer)
 
 ### users.list [GET /users.list]
 
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -41,7 +41,7 @@ Get the current authenticated user.
             + preferences (object)
                 + invoiceable: true (boolean)
                 + time_tracking_freezing: true (boolean)
-                + time_tracking_freezing_cutoff_days: 7 (integer)
+                + time_tracking_freezing_cutoff_days: 7 (number)
 
 ### users.list [GET /users.list]
 
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -43,8 +43,8 @@ Get the current authenticated user.
                 + historic_time_tracking_limit (object, nullable)
                     + unit (enum)
                         + Members
-                            + day - Days
-                    + value: 7 (number)
+                            + hour - Hours
+                    + value: 24 (number)
 
 ### users.list [GET /users.list]
 
--- a/src/01-general/users.apib
+++ b/src/01-general/users.apib
@@ -40,8 +40,11 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
-                + time_tracking_freezing: true (boolean)
-                + time_tracking_freezing_cutoff_days: 7 (number)
+                + historic_time_tracking_limit (object, nullable)
+                    + unit (enum)
+                        + Members
+                            + day - Days
+                    + value: 7 (number)
 
 ### users.list [GET /users.list]
 
--- a/src/changes-backwards-compatible.apib
+++ b/src/changes-backwards-compatible.apib
@@ -6,6 +6,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We exposed the "Freeze time tracking after a number of days" preference in `users.me`.
+
 - We added format `ubl/e-fff` to `invoices.download`.
 
 - We added filters `estimated_closing_date_from` and `estimated_closing_date_until` to `deals.list`.


src/01-general/users.apib
```diff
@@ -40,6 +40,8 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
+                + time_tracking_freezing: true (boolean)
+                + time_tracking_freezing_cutoff_days: 7 (number)
```
dietrichm: Q: what if we (in the future) want to freeze time trackings after i.e. 8 hours? How can we reflect this in the `users.me` response?

src/01-general/users.apib
```diff
@@ -40,6 +40,8 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
+                + time_tracking_freezing: true (boolean)
```
dietrichm: Q: can we make this one (object) property instead of two related ones?

stivni:
I don't see how you're going to use this?
Shouldn't this be a back-end concern?

adognameddog:
> I don't see how you're going to use this?
> Shouldn't this be a back-end concern?

It is a back-end concern. But front-end, and that includes mobile, needs to know wether to disable the editing of a time tracking record. So they need to know if this setting/preference is enabled, and how long the period is.

That settings needs to be exposed in the API. If we were to inject it, with an _includes_, into the time tracking api endpoints, we need to introduce a new interface, and two new traits into the code of that repository. So we opt for adding it to `users.me`

src/01-general/users.apib
adognameddog: I also would go for a value/unit object here.

src/01-general/users.apib
adognameddog: Yup, I would go for time_tracking_freezing -> enabled, period (optional). And period would be a value/unit object.

stivni:
> So they need to know if this setting/preference is enabled, and how long the period is.

I don't think so. 

They (just?) need to know if a time tracking record is locked or not (for whatever reason, unless you want to show more info about it, but I would avoid that at this moment).

Is that what you mean with 

> That settings needs to be exposed in the API. If we were to inject it, with an `includes`, into the time tracking api endpoints, we need to introduce a new interface, and two new traits into the code of that repository. So we opt for adding it to `users.me`

?

adognameddog:
> > So they need to know if this setting/preference is enabled, and how long the period is.
> 
> I don't think so.
> 
> They (just?) need to know if a time tracking record is locked or not (for whatever reason, unless you want to show more info about it, but I would avoid that at this moment).
> 
> Is that what you mean with
> 
> > That settings needs to be exposed in the API. If we were to inject it, with an `includes`, into the time tracking api endpoints, we need to introduce a new interface, and two new traits into the code of that repository. So we opt for adding it to `users.me`
> 
> ?

That's also an option I guess 👍 We will discuss it

stivni:
> We will discuss it

I hope so, because I don't think I can support this approach 😅

ketlane:
Aside from locking a time tracking from being updated, the UI also needs to be able to prevent a user from adding a time tracking in the past. 

We could say "ok, the backend will respond with a 400 error" in such cases, after receiving the request to save a time tracking in the past; and the UI should translate that backend error and display it. 

But it would be more user friendly to have this validation in the front end as well, preventing the user from being able to click the "Add/Save" button immediately as the UI notices the date is wrong. Or even limit the date picker to the last 7 days only, if that's what the account administrator decided.

stivni:
> the UI also needs to be able to prevent a user from adding a time tracking in the past.

Aha! That's a use case I didn't know about yet.

In that case, it could indeed be a preference, although I'm not sure if it's on the user level.
Can there be a difference between different users in one account?

stivni:
I would however do a "locked/editable" concept on the record level anyways.
And also for the `add`, `update`, `delete`, `timers.stop` do back-end validation.

For the front-end UX, we can iterate on this based on my question above.

ketlane:
> Can there be a difference between different users in one account?

Yes, it's an account level preference (just like the `invoiceble` preference in users.me is, btw). But we don't have an account settings endpoint.

src/01-general/users.apib
JordyDeGroeve: +1 ☝️

stivni:
For `invoiceble` we made the explicit choice to bring it immediately on the user level, because it makes sense that users (or their manager) could toggle this individually.

I wonder if there is an (obvious) use-case for letting this setting differ on a per user base?

Anyhow, I know it's called "freezing" in the UI. But is that concept actually clear? It is only mentioned briefly in the KB, in Dutch it's called "bevriezen" in the UI, but "blokkeren" in the knowledge base. I would personally call it "locking" (and eventually remove the timebased component of it and let admins lock and unlock at will). (<- all optional in the discussion btw)

I wonder if we need two fields for this... 
Maybe something like `historic_time_tracking_limit: { unit: days, value: 7} || null` might be enough?

ketlane:
> I wonder if there is an (obvious) use-case for letting this setting differ on a per user base?

It's currently the case that regular users are affected by this setting, but not administrators; but this is not clearly marked anywhere in the UI. I think it may just be that in time we will decide to make this explicit and transform this account preference into user permission, instead, thus making it explicit who is affected by the limitation. 
 
> I wonder if we need two fields for this...
> Maybe something like `historic_time_tracking_limit: { unit: days, value: 7} || null` might be enough?

I'm ok with that.

stivni:
> > I wonder if there is an (obvious) use-case for letting this setting differ on a per user base?
> 
> It's currently the case that regular users are affected by this setting, but not administrators; but this is not clearly marked anywhere in the UI. I think it may just be that in time we will decide to make this explicit and transform this account preference into user permission, instead, thus making it explicit who is affected by the limitation.
> 
> > I wonder if we need two fields for this...
> > Maybe something like `historic_time_tracking_limit: { unit: days, value: 7} || null` might be enough?
> 
> I'm ok with that.

+1 in that case

JordyDeGroeve (dismissed):

apiary.apib
```diff
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We exposed the "Freeze time tracking after a number of days" preference in `users.me`.
```
JordyDeGroeve, line 377: Q: do we need to update the description to resemble the updated name?

ketlane: That's how it's called in the UI currently.

ketlane: image

JordyDeGroeve: Okay, clear enough than. 👍

tiagobastos (dismissed):

src/01-general/users.apib
```diff
@@ -40,8 +40,11 @@ Get the current authenticated user.
             + time_zone: `Europe/Brussels` (string)
             + preferences (object)
                 + invoiceable: true (boolean)
-                + time_tracking_freezing: true (boolean)
-                + time_tracking_freezing_cutoff_days: 7 (number)
+                + historic_time_tracking_limit (object, nullable)
+                    + unit (enum)
+                        + Members
+                            + day - Days
+                    + value: 7 (number)
```
dietrichm: PP: use `hours` as the basic unit here. An hour seems like exactly the right granularity and doesn't require us to add a new unit once we want to limit to i.e. 8 hours.

dietrichm (dismissed):

mikeverf (dismissed):

src/01-general/users.apib
ketlane: The disadvantage of always using hours is that it doesn't directly map to the value added in the UI, which may induce some degree of uncertainty.

On the other hand, if we ever decide to introduce a different granularity (i.e. hours), having the unit here means we can also return hours and change the unit to hours at that point.

But I will change the default unit to hours if more people feel this is the right way to go.

ketlane:
> I would however do a "locked/editable" concept on the record level anyways.

Regarding this: it depends on who you are. If you are an administrator, you can modify it anyway. Also, if you are a team manager, you can modify entries for your subordinates.

So front-ends/integrations would still need to decide based on the current user (and possibly on their relationship with the time tracking user, when managing timesheets for other users), not only on the `locked` property. That would make such a field difficult to interpret in the API response. 

I guess we could have the API populate a `locked` field not only on time + lock preference, but also based on who the current user is (and their relationship with the time tracking user). That would make the `timeTracking.info` and `timeTracking.list` calls not cacheable, though: I'm not sure whether this is the right direction.

src/01-general/users.apib
dietrichm: We can definitely add more units in the future. However residing to hours from the start will give us a lot of flexibility for the future, i.e. not having to update the mobile app if we allow the cut-off to be less than a day.

I'd even dare to say we could then ever only return the value in hours and not require any integrator to recalculate to a different unit.

dietrichm:
> I guess we could have the API populate a `locked` field not only on time + lock preference, but also based on who the current user is (and their relationship with the time tracking user). That would make the `timeTracking.info` and `timeTracking.list` calls not cacheable, though: I'm not sure whether this is the right direction.

Good call. We could also have two distinct properties, one `locked` which denotes the time tracking is locked (i.e. the owner cannot edit it anymore) and a more meta level property denoting whether the _authenticated user_ can edit it.

I'm not entirely sure if this is a right approach. It feels like adding front-end logic to the API, although it *is* the API deciding whether the authenticated user can edit the time tracking (so no front-end or integration has to).

What do you think about this?

stivni:
> it is the API deciding whether the authenticated user can edit the time tracking

exactly

adognameddog (dismissed):

src/01-general/users.apib
ketlane: Ok, changed to hours

dietrichm (approved):

tiagobastos (approved):

JordyDeGroeve (approved):

mikeverf (approved):

ketlane:
I moved the `locked` / `updatable` discussion to a different PR (https://github.com/teamleadercrm/api/pull/498), since it's not strictly tied to this one.

adognameddog (approved):

mikeverf (approved):

PR #501, opened 2020-02-27
New filter "relates_to" on timeTrackings.list [PRO-3401]

adognameddog:
If you filter on "relates to milestone", all time tracked on tasks, meetings, calls, tickets belonging to that milestone will be returned, as well as the time tracked directly on that milestone

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3683,6 +3683,11 @@ Get a list of tracked time.
                         + milestone
                         + ticket
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + relates_to (object, optional) - Find all tracked time linked directly and indirectly to a subject
+                + type (enum, required)
+                    + Members
+                        + milestone
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
         + page (Page, optional)
         + sort (array, optional)
             + (object)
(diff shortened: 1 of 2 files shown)


mikeverf (approved):

apiary.apib
```diff
@@ -3683,6 +3683,11 @@ Get a list of tracked time.
                         + milestone
                         + ticket
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + relates_to (object, optional) - Find all tracked time linked directly and indirectly to a subject
+                + type (enum, required)
+                    + Members
+                        + milestone
```
stivni, line 3689: Is it a lot of work to support this for other types too?
And would we benefit from making this an array immediately?

joaoinacio: > And would we benefit from making this an array immediately?

Would that make elements part of an `AND` or `OR` filter ? 🤔

ArnaudWeyts (approved):

apiary.apib
adognameddog: @stivni We are going to implement it for `project` very soon, since we are going to need that. So I made it an array (also for more consistency)

adognameddog: And it should not be an array. You provide one relation in the filter. At least that's for now. Combining relations is YAGNI

ArnaudWeyts:
✅ Manually checked

joaoinacio (approved):

PR #502, opened 2020-03-02
Filter tracked time on subject type [PRO-3411]

adognameddog:
With an array that may contain one or more elements. NULL means no subject type.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3701,6 +3701,14 @@ Get a list of tracked time.
                         + milestone
                         + ticket
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + subject_types (array, optional) - Include all tracked time with one of the provided subject types. For tracked time without a subject type, provide `null`
+                + Members
+                    + company
+                    + contact
+                    + event
+                    + todo
+                    + milestone
+                    + ticket
             + relates_to (object, optional) - Find all tracked time linked directly and indirectly to a subject
                 + type (enum, required)
                     + Members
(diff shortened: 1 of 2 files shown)


tiagobastos (approved):

apiary.apib
```diff
@@ -3701,6 +3701,14 @@ Get a list of tracked time.
                         + milestone
                         + ticket
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + subject_types (array, optional) - Include all tracked time with one of the provided subject types. For tracked time without a subject type, provide `null`
+                + Members
+                    + company
+                    + contact
+                    + event
+                    + todo
```
mrtus, line 3709: Q: Isn't `todo` a deprecated term? Shouldn't this be `task`?

adognameddog: In our API, it's still `todo`. We should change that to task, but that's very much a breaking change, witch requires a new version, and is out of scope for the moment.

mikeverf (approved):

apiary.apib
joaoinacio: 🤔 I seem to recall some endpoints (as well as sideloader?) moving to `task` whenever possible while still supporting legacy `todo` when required..
Will try to find some examples.

**Edit**:
SubjectType is not one of them, so `todo` still seems pretty much required...

joaoinacio (approved):

ketlane (approved):

ArnaudWeyts (approved):

ArnaudWeyts:
✅ MC'd

PR #503, opened 2020-03-03
Added Duration datastructure

adognameddog:
```
## Duration (object)
+ unit: `s` (string, required) - Seconds
+ value: `10800` (number, required)
```

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -4309,9 +4309,5 @@ Unregister a webhook.
                 + `payment_card`
 
 ## Duration (object)
-+ unit: `min` (enum[string], required)
-    + Members
-        + `h` - Hours
-        + `min` - Minutes
-        + `s` - Seconds
-+ value: `180` (number, required)
++ unit: `s` (string, required) - Seconds
++ value: `10800` (number, required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -4309,5 +4309,7 @@ Unregister a webhook.
                 + `payment_card`
 
 ## Duration (object)
-+ unit: `s` (string, required) - Seconds
++ unit: `s` (enum[string], required)
+    + Members
+        + `s` - Seconds
 + value: `10800` (number, required)
--- a/apiary.apib
+++ b/apiary.apib
@@ -4307,3 +4307,11 @@ Unregister a webhook.
                 + `bankers_draft`
                 + `bank_transfer`
                 + `payment_card`
+
+## Duration (object)
++ unit: `min` (enum[string], required)
+    + Members
+        + `h` - Hours
+        + `min` - Minutes
+        + `s` - Seconds
++ value: `180` (number, required)
--- a/src/datastructures.apib
+++ b/src/datastructures.apib
@@ -162,9 +162,5 @@
                 + `payment_card`
 
 ## Duration (object)
-+ unit: `min` (enum[string], required)
-    + Members
-        + `h` - Hours
-        + `min` - Minutes
-        + `s` - Seconds
-+ value: `180` (number, required)
++ unit: `s` (string, required) - Seconds
++ value: `10800` (number, required)
--- a/src/datastructures.apib
+++ b/src/datastructures.apib
@@ -160,3 +160,11 @@
                 + `bankers_draft`
                 + `bank_transfer`
                 + `payment_card`
+
+## Duration (object)
++ unit: `min` (enum[string], required)
+    + Members
+        + `h` - Hours
+        + `min` - Minutes
+        + `s` - Seconds
++ value: `180` (number, required)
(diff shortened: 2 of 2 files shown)


JordyDeGroeve (dismissed):
Will it be used in the near future?

adognameddog:
> Will it be used in the near future?

I don't know yet. Debating wether to use in on `timeTracking.report`, but I don't know if we want to suggest that we might return a duration with, suddenly, a new unit

JordyDeGroeve:
Not sure if this should be added if there is no concrete use case for it.
We can leave out the other units perhaps?
What does @stivni think?

adognameddog:
> Not sure if this should be added if there is no concrete use case for it.
> We can leave out the other units perhaps?
> What does @stivni think?

I'm going to use it in the docs for `timeTracking.report`

JordyDeGroeve (dismissed):

ArnaudWeyts (dismissed):

tiagobastos (dismissed):

apiary.apib
```diff
@@ -4307,3 +4307,7 @@ Unregister a webhook.
                 + `bankers_draft`
                 + `bank_transfer`
                 + `payment_card`
+
+## Duration (object)
++ unit: `s` (string, required) - Seconds
```
tiagobastos: Q: Why not do it like: 
```
+ unit: `s` (enum[string], required)
    + Members
        + s - Seconds
```

We _"did it"_ for Distance on Work Orders.
https://github.com/teamleadercrm/api-internal/pull/60/files#diff-1841f13b750b4c8d06587142329fe64cR177-R181

mikeverf (dismissed):

apiary.apib
joaoinacio: `DurationUnit` also provides minutes, as `min` so that would actually make it:
```
+ unit: `s` (enum[string], required)
    + Members
        + s - Seconds
		+ min - Minutes
```

joaoinacio: We already use duration in minutes on some endpoints, so having this support it would be ideal IMO.

tiagobastos: And not in the scope of this proposal to change, but to provide more insight, the `/tasks.create` already uses a duration like structure on `estimated_duration`. 
That could be a good example of it's usage already.

dietrichm: Agreeing with @joaoinacio and @tiagobastos. Let's define `unit` as an enum so we are not fixed to having just one possible unit.

dietrichm (changes requested):

joaoinacio (approved):

ketlane (commented):
The already existing documentation uses another notation when expressing duration units: `min` for minutes (i.e. on `tasks.list` for `estimated_duration`). That would mean for seconds we'd have to use `sec` not `s`, if we want to be consistent.  Otherwise, changing `min` to `m` in already implemented / documented endpoints would imply a breaking change.

So I'm proposing to use `sec` instead of `s`.

JordyDeGroeve (approved):

joaoinacio:
> The already existing documentation uses another notation when expressing duration units: `min` for minutes (i.e. on `tasks.list` for `estimated_duration`). That would mean for seconds we'd have to use `sec` not `s`, if we want to be consistent. Otherwise, changing `min` to `m` in already implemented / documented endpoints would imply a breaking change.
> 
> So I'm proposing to use `sec` instead of `s`.

However there is also already existing documentation using `s` notation for seconds.
This really should be standardised across existing endpoints, but either way it would/will involve some kind of (breaking?) change.

Also, [DurationUnit](https://github.com/teamleadercrm/core/blob/master/src/Tasks/Application/Http/DurationUnit.php#L17) does use `s` and `min`... 🤷‍♂

tiagobastos:
Should we decide on the implementation we would like to see in all the endpoints?
And the ones that differ would be adjusted in the context of each team. 
> So I'm proposing to use sec instead of s.

Like this approach too.

adognameddog:
> The already existing documentation uses another notation when expressing duration units: `min` for minutes (i.e. on `tasks.list` for `estimated_duration`). That would mean for seconds we'd have to use `sec` not `s`, if we want to be consistent. Otherwise, changing `min` to `m` in already implemented / documented endpoints would imply a breaking change.
> 
> So I'm proposing to use `sec` instead of `s`.

Absolutely not. The international standard abbreviation for seconds is "s". See: https://maths.mq.edu.au/numeracy/tutorial/siunits.htm

We use "min" for minutes because minutes is not a SI unit, and therefor we can practically chose the abbreviation. "m" is metres.

ArnaudWeyts (approved):

ketlane (approved):

tiagobastos (approved):

mikeverf (approved):

dietrichm (approved):
Absolutely in favour of using SI units (`s` for seconds).

PR #504, opened 2020-03-03
timeTracking.report [PRO-3414]

adognameddog:


adognameddog:
Something I'm not sure of: should `count` be on the same level as `totals`, or as it is now: on the same level as `tracked_time`, `invoiceable`...

apiary.apib
```diff
@@ -3786,6 +3786,70 @@ Get information about tracked time.
...
+                + type (enum, required)
+                    + Members
+                        + company
+                        + contact
+                        + event
+                        + todo
+                        + milestone
+                        + ticket
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + subject_types (array, optional) - Include all tracked time with one of the provided subject types. For tracked time without a subject type, provide `null`
```
stivni, line 3812: Q : shouldn't we include `null` as a possible member?

adognameddog: Ah yes

apiary.apib
```diff
@@ -3786,6 +3786,70 @@ Get information about tracked time.
...
+            + relates_to (object, optional) - Find all tracked time linked directly and indirectly to a subject
+                + type (enum, required)
+                    + Members
+                        + milestone
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+            + `grouped_by`: `subject` (enum[string], required) - Return the totals of the filtered tracked time, grouped by this value
+                + Members
+                    + subject
+                    + `subject_type`
+                    + `related_milestone`
```
stivni, line 3829: Q : it's weird that `subject` is an existing property, a `subject_type` is an existing property, but one level deeper (a subject's type) but `related_milestone` seems like a different concept... 

I understand your usecase to find all trackings for milestones, but I would like to think a bit about how to do this in practice in a more clear and composable way...

apiary.apib
```diff
@@ -3786,6 +3786,70 @@ Get information about tracked time.
...
+                    + company
+                    + contact
+                    + event
+                    + todo
+                    + milestone
+                    + ticket
+            + relates_to (object, optional) - Find all tracked time linked directly and indirectly to a subject
+                + type (enum, required)
+                    + Members
+                        + milestone
```
stivni, line 3823: Q : Why don't we include other types here?

adognameddog: A: In the future, we might add more, but each entity requires very specific logic. For now, we only allow related to milestone.

apiary.apib
adognameddog: `subject` means grouped by each individual subject
`subject_type` means grouped by subject type
`related_milestone` means grouped by not only tracked on an individual milestone, but also tracked on something related to an individual milestone.

It's maybe a good idea to add some more explanation there

apiary.apib
stivni: Really? Can you explain me irl later today?

adognameddog:
I'm moving this to our internal-api documentation to further weed out the needs and wants of this endpoint.

PR #505, opened 2020-03-03
Work Orders | Un-publish endpoint

tiagobastos:
Q: Is the explanation on the changelog enough like it is?


JordyDeGroeve (dismissed):
PP: add some extra context or a reason why we did it.

kevinwaelkens (dismissed):

nielsvermaut (dismissed):

nielsvermaut:
@stivni Tiago suggested that you would merge this if this looks good to you.

src/changes-backwards-compatible.apib
```diff
@@ -6,6 +6,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We removed all `Work Orders` documentation.
```
stivni: PP : please don't mention this in the changelog.

The reason we would unpublish these docs is to remove focus from this, since the group of endpoints is not complete. No need to stress that out

src/changes-backwards-compatible.apib
```diff
@@ -6,6 +6,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We removed all `Work Orders` documentation. 
+
```
stivni: PP : please don't mention this in the changelog.

The reason we would unpublish these docs is to remove focus from this, since the group of endpoints is not complete. No need to stress that out

```suggestion
```

JordyDeGroeve (approved):

steftrenson (approved):

ketlane (approved):

PR #513, opened 2020-03-25
PRO-3582: document `dependency_for` in milestones.info and .list

joaoinacio:


ketlane (dismissed):

mikeverf (dismissed):

apiary.apib
```diff
@@ -399,6 +399,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 - We added `depends_on` to the `milestones.info` and `milestones.list` endpoints.
 
+- We added `dependency_for` to the `milestones.info` and `milestones.list` endpoints. This indicates all the milestones for which the current milestone is a dependecy.
```
adognameddog: Typo: `dependecy`

adognameddog (dismissed):

apiary.apib
```diff
@@ -3342,6 +3344,10 @@ Get a list of project milestones.
                 + depends_on (object, nullable)
                     + type: `milestone` (string)
                     + id: `0488d792-ba9b-059f-bd57-bea75d3f4f4e` (string)
+                + dependency_for (array)
```
stivni, line 3347: Is it really necessary to have the relationship modelled in two directions?
Where will we use this?

adognameddog (dismissed):

apiary.apib
adognameddog: Yes. The frontend will tell our users if their changes to due dates (or dependencies) will mess up the project's flow. That goes for both the "parent" and "children". An alternative would be to create a filter, but that would make it all more complex, and would mean more API calls

ArnaudWeyts (dismissed):

adognameddog (approved):

ArnaudWeyts (approved):

mikeverf (approved):

PR #516, opened 2020-04-01
Expose the milestone description

ketlane:
### What has been done
- Added support for extracting the milestone description in `milestones.info`
- Added support for setting a milestone description in `milestones.create` and `milestones.update`

### DIFF
--- a/src/07-projects/milestones.apib
+++ b/src/07-projects/milestones.apib
@@ -85,6 +85,7 @@ Get details for a single milestone.
                 + id: `fffaf951-b65a-4cb1-bfb6-4b93e0e60c10` (string)
             + due_on: `2018-01-01` (string)
             + name: `Initial setup` (string)
+            + description (string)
             + responsible_user (object)
                 + type: `user` (string)
                 + id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string)
@@ -115,6 +116,7 @@ Create a new milestone.
         + project_id: `1c159f98-4b07-438a-9f42-fb4206b9530d` (string, required)
         + due_on: `2018-01-01` (string, required)
         + name: `Initial setup` (string, required)
+        + description (string, optional)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, required)
         + One Of
             + Properties
@@ -147,6 +149,7 @@ Update a milestone.
         + id: `67e80ad8-d14f-4510-a2bd-a4c6aa578c37` (string, required)
         + due_on: `2018-01-01` (string, optional)
         + name: `Initial setup` (string, optional)
+        + description (string, optional)
         + responsible_user_id: `e1240972-6cfc-4549-b49c-edda7568cc48` (string, optional)
         + depends_on: `0488d792-ba9b-059f-bd57-bea75d3f4f4e` (string, nullable)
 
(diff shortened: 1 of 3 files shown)


mrtus (approved):

luisaterrabrito (approved):

timdegroote (approved):

tiagobastos (approved):

lowiebenoot (approved):

kevinwaelkens (approved):

adognameddog (approved):

src/07-projects/milestones.apib
```diff
@@ -85,6 +85,7 @@ Get details for a single milestone.
                 + id: `fffaf951-b65a-4cb1-bfb6-4b93e0e60c10` (string)
             + due_on: `2018-01-01` (string)
             + name: `Initial setup` (string)
+            + description (string)
```
dietrichm, line 88: Q: if the description is optional when creating a milestone, does it get a value implicitly? Otherwise we should make this property nullable.

src/07-projects/milestones.apib
```diff
@@ -147,6 +149,7 @@ Update a milestone.
         + id: `67e80ad8-d14f-4510-a2bd-a4c6aa578c37` (string, required)
         + due_on: `2018-01-01` (string, optional)
         + name: `Initial setup` (string, optional)
+        + description (string, optional)
```
dietrichm, line 152: Q: kinda the same question here. If we want to unset a description of a milestone, we need to be able to pass `null` as a value, hence the property should be nullable.

Not passing it would equal not updating it.

src/07-projects/milestones.apib
ketlane: It's empty by default

src/07-projects/milestones.apib
ketlane: It's possible to unset the description by passing an empty string.

joaoinacio: > It's possible to unset the description by passing an empty string.

More correctly, the description cannot be unset, but it can be empty (string).

joaoinacio (approved):

src/07-projects/milestones.apib
dietrichm: Is there a difference between an empty string (which technically is a value) and no value (`null`) for a milestone's description in the domain? If there is no difference, I'd treat the empty string as `null`.

jnstr (approved):

src/07-projects/milestones.apib
dietrichm: IIRC, we do have endpoints that treat `''` as `null` and unset the property. Cfr the other comment as well.

src/07-projects/milestones.apib
joaoinacio: In domain, there is no `null`able description. Neither in DB.
So, there is a difference I guess? (an empty description would always be returned as an empty string, never `null`)

ketlane: We do not support nulls in this field. I prefer this to be transparent throughout the application and the API.

dietrichm: I do think a non nullable column in the database in this case is rather a design error which should not be reflected in the domain or public API.

Let's settle on the current proposal but take this into account in the future. 😃

PR #517, opened 2020-04-01
[IN-2691] Pass the name of the quotation info

EOSullivanBerlin:
We need to be consistent in displaying the names of quotations between the history and other places that we show the quotation. 
Having the backend take care of this ensures constancy between languages.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1997,6 +1997,7 @@ Get a quotation
     + Attributes (object)
         + data (object)
             + id: `e7a3fe2b-2c75-480f-87b9-121816b5257b` (string)
+            + name: `Quotation 1234` (string)
             + deal (object)
                 + type: `deal` (string)
                 + id: `53474a7a-f9b2-4dd4-88a8-40ce773c7a64` (string)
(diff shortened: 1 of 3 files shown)


luisaterrabrito (dismissed):

steftrenson (dismissed):

dietrichm (dismissed):

lowiebenoot (dismissed):
Q: is the filename currently always the same as the name?

sanderbrugge (dismissed):

apiary.apib
```diff
@@ -1997,6 +1999,7 @@ Get a quotation
     + Attributes (object)
         + data (object)
             + id: `e7a3fe2b-2c75-480f-87b9-121816b5257b` (string)
+            + name: `Quotation 1234` (string)
```
mrtus: Q: For deals I see we call this a `title`, would it make sense to keep the same naming here too?

What do you think?

EOSullivanBerlin:
@lowiebenoot I agree that they should be consistent but I don't know.

apiary.apib
dietrichm: Good remark. I vote for consistency here.

EOSullivanBerlin: I will change it to title.

sanderbrugge (approved):

mrtus (approved):

luisaterrabrito (approved):

dietrichm (approved):

lowiebenoot (approved):

steftrenson (approved):

PR #518, opened 2020-04-08
Add the file to the quotation info response 

EOSullivanBerlin:
This will allow us to show the preview of the quotation in the follow up section of the deals detail.

Is this a good approach?


lowiebenoot (approved):

src/03-deals/quotations.apib
```diff
@@ -21,6 +21,9 @@ Get a quotation
             + deal (object)
                 + type: `deal` (string)
                 + id: `53474a7a-f9b2-4dd4-88a8-40ce773c7a64` (string)
+            + file (object)
+                + type: `file` (string)
+                + id: `d8c47d9f-b342-055d-961d-daea196256a1` (string)
```
dietrichm, line 26: Not entirely convinced about this yet.

`file` is a bit arbitrary: what does it refer to exactly? It could be a user uploaded file, it could be the PDF generated by Teamleader, or it could even refer to a different representation of the quotation (in a different format than PDF).

I'm thinking an extended version of a download endpoint (like it's available for work orders) could work and might be the preferred solution (especially since it already takes a format property i.e. `pdf`), but will require more work and will also require additional API calls in the front end.

Considering the use case for this I'd check if we can let the back end decide which file to show given the entity UUID. Then we don't need to account this for in the API.

EOSullivanBerlin: Okay I'm happy to try and expand the download endpoint. 
The extra API calls would be a pity but if thats the only way to do it then lets go with that.

dietrichm: I would not create more download endpoints just yet. Hence the idea for fixing this in the back end. Let's have a TA about this.

PR #519, opened 2020-04-14
IN-2460 Add margin to deals.info response

EOSullivanBerlin:
I made it nullable because we will not always have a quotation and therefore no profit margin or should I just pass zero?


EOSullivanBerlin:
Closing because it should be internal.

PR #526, opened 2020-05-11
PRO-3623 - Add purchase_order_number to projects

Pegasuz:
Do not merge before implementation is merged!

This PR adds the "purchase_order_number" to the projects.info, projects.create and projects.update endpoints.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,6 +374,10 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added the `purchase_order_number` field to `projects.info`, `projects.create`, and `projects.update`
+
+#### April 2020
+
 - We added the `description` field to `milestones.info`, `milestones.create`, and `milestones.update`
 
 - We added the `propagate_date_changes` property to `milestones.update`. This allows propagating the new due date difference to the due date of all open tasks of that milestone, and recursively to all depending milestones.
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -3201,6 +3205,7 @@ Get details for a single project.
                         + Members
                             + decision_maker
                             + member
+            + purchase_order_number: `000023` (string)
```
joaoinacio: It should be `nullable`, IIRC?

adognameddog (changes requested):
Purchase order number should be nullable, like @joaoinacio mentioned. Otherwise looking good!

ketlane (dismissed):

apiary.apib
```diff
@@ -374,6 +374,10 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added the `purchase_order_number` field to `projects.info`, `projects.create`, and `projects.update`
+
+#### April 2020
```
ketlane, line 379: Not sure we should add the month grouping in this PR, but I'm not sure we shouldn't either. Maybe there are other opinions about it?

ArnaudWeyts (dismissed):

apiary.apib
adognameddog: If someone adds new documentation in June, then that person should add a new header and place everything from latest in there.

adognameddog (dismissed):

mikeverf (dismissed):

joaoinacio (dismissed):

adognameddog (approved):

mikeverf (approved):

ArnaudWeyts (approved):

PR #529, opened 2020-05-26
Remove unused argument for timers.stop

JordyDeGroeve:
Remove an argument that is never used.

The `ended_at` time is never used in the implementation.
It won't be tackled any time soon so we will remove it.

This is **no breaking change** since the request will still work (with the ignored argument) and the behaviour stays the same because it wasn't used before.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -4071,11 +4071,6 @@ Start a new timer.
 
 Stop the current timer. This will add a new time tracking in the background.
 
-+ Request (application/json)
-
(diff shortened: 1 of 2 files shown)


sanderbrugge (approved):

kevinwaelkens (approved):

tiagobastos (approved):

nielsvermaut (approved):

tiagobastos (approved):
✅ Correctly removed

PR #530, opened 2020-06-03
Added a header for May 2020

adognameddog:
What it says on the tin

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -372,7 +372,7 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
-#### Latest
+#### May 2020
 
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -372,7 +372,7 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
-#### Latest
+#### May 2020
```
luisaterrabrito, line 375: Can you add a new latest?

dietrichm (approved):

mrtus (approved):

apiary.apib
adognameddog: I could, but there would be nothing in there?

luisaterrabrito (approved):

apiary.apib
luisaterrabrito: Yeah let's add it again when there are new changes

ketlane (approved):

dietrichm:
✅ Checked

PR #532, opened 2020-06-04
[PROPOSAL] Addresses | Area level two

tiagobastos:
Proposing to add _**Area Level Two**_ on the `Address` data structure.

Initial the plan was to roll it out for customer but because this is part of the address I opted for adding it to the `Address` object as a nullable property. 

The places affected by the change would be:
- Contacts
- Companies
- Departments 
- Work Orders

Objections?
Opinions?

### DIFF
--- a/src/datastructures.apib
+++ b/src/datastructures.apib
@@ -20,6 +20,9 @@
 + postal_code: `9000` (string, required, nullable)
 + city: `Ghent` (string, required, nullable)
 + country: `BE` (string, required)
++ area_level_two (object, nullable)
+    + type: `area_level_two` (string)
+    + id: `fd48d4a3-b9dc-4eac-8071-5889c9f21e5d` (string)
 
 ## Addressee (object)
 + addressee: `Teamleader HQ` (string, optional)
--- a/apiary.apib
+++ b/apiary.apib
@@ -4258,6 +4258,9 @@ Unregister a webhook.
 + postal_code: `9000` (string, required, nullable)
 + city: `Ghent` (string, required, nullable)
 + country: `BE` (string, required)
++ area_level_two (object, nullable)
+    + type: `area_level_two` (string)
+    + id: `fd48d4a3-b9dc-4eac-8071-5889c9f21e5d` (string)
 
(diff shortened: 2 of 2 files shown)


src/datastructures.apib
```diff
@@ -20,6 +20,9 @@
 + postal_code: `9000` (string, required, nullable)
 + city: `Ghent` (string, required, nullable)
 + country: `BE` (string, required)
++ area_level_two (object, nullable)
```
tiagobastos, line 23: Should it be `nullable` and always show. 
Or `optional` and only be given when set?

kevinwaelkens (approved):

src/datastructures.apib
sanderbrugge: My pref is nullable and always shown.

sanderbrugge (approved):

src/datastructures.apib
JordyDeGroeve: > My pref is nullable and always shown.

Same.

JordyDeGroeve (approved):

src/datastructures.apib
tiagobastos: Asked because it's already a pattern on the address.
`addressee` under `Addressee` is already only shown when set. 
As a PP, I do prefer `nullable` too.

luisaterrabrito (approved):

nielsvermaut (approved):

adognameddog (approved):

joaoinacio (approved):

JordyDeGroeve:
Let's wait for merging until the area level two feature is rolled out.

dietrichm (approved):

nielsvermaut:
@tiagobastos We've released this, so it can be merged, right?

tiagobastos:
> @tiagobastos We've released this, so it can be merged, right?

No,this proposal need's to be updated.
We need to duplicate the data structure that's used both in the read and write side.

JordyDeGroeve:
Closed in favour of https://github.com/teamleadercrm/api/pull/573

PR #534, opened 2020-06-08
[PRO-3950] Do not require all fields on project.update endpoint

Pegasuz:
We no longer require all fields to be present on the `projects.update` API endpoint.
Only the `id`is required, all other fields are optional.

Do not merge untill implementation is ready.

### DIFF
--- a/src/07-projects/projects.apib
+++ b/src/07-projects/projects.apib
@@ -196,17 +196,17 @@ Update a project.
 
     + Attributes (object)
         + id: `dcc2e8ed-51be-4cb6-9c01-034aedac86fd` (string, required)
-        + title: `New company website` (string, required)
-        + description (string, required)
-        + status: `active` (enum, required)
+        + title: `New company website` (string)
+        + description (string)
+        + status: `active` (enum)
             + Members
                 + active
                 + on_hold
                 + done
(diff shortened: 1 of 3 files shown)


joaoinacio (approved):

ketlane (approved):

src/07-projects/projects.apib
```diff
@@ -196,17 +196,17 @@ Update a project.
 
     + Attributes (object)
         + id: `dcc2e8ed-51be-4cb6-9c01-034aedac86fd` (string, required)
-        + title: `New company website` (string, required)
-        + description (string, required)
-        + status: `active` (enum, required)
+        + title: `New company website` (string)
+        + description (string)
```
adognameddog, line 200: Q: isn't this nullable? So when it's null, the description is removed?

joaoinacio: Not part of the scope, AFAICT.

It shouldn't be too hard to implement it (the Update command seems to already support it), but passing an empty string would work just as well so.. 🤷

adognameddog (approved):

ArnaudWeyts (approved):

mikeverf (approved):

PR #535, opened 2020-06-08
Added companies to contacts.list response

Brammm:
It's currently not possible to get the company(ies) a contact is related to in the contacts.list endpoint.


mrtus (commented):
Hello @Brammm 

Thank you for making this request.
As you might notice on our `.list` endpoints we do not always return the same information as we do in our `.info`. We do this on purpose to keep our `.list` endpoints reasonable in size and speedy.

We do want to note that this field is available on the `.info` endpoint.

Brammm:
Hey @mrtus 

The list endpoint is paginated as it is, so I assumed performance would be less of a factor. As it stands, we are forced to call the .info endpoint for a collection of contacts, because we require the company info, which is a bunch more calls than we had hoped.

Would an optional include e.g. `&includes[]=company_info` or something like that be an option? We'd happily take a performance hit on the list endpoint over having to call .info multiple times.

justijndepover:
Hi, 

Is this a feature you guys are looking in to?
I'm having the same problem, where I need to do a separate call to the .info endpoint, just to get the company relation.

In my case this will result in +- 15000 extra api calls.

Brammm:
@justijndepover yawza. Make sure to build in some delays between requests or you're gonna be hit with rate limits as well...

justijndepover:
Hi @mrtus 

I'd like to point out that this remains an issue for me (and probably for others as well).
Do you guys plan on a solution for this?

As @Brammm mentioned, I too would prefer a performance hit instead of making thousand of extra requests.
(Which, in the end, would result in an overal slower process)

marievandaele:
Feel free to send an email to support@teamleader.eu, asking to be linked  to issue ID-4236 (see related companies in return data of contacts.list). That way we can update you once this has been added to our API

AxelDeneu:
Hello,
Sorry in advance to dig up this issue but is this planned to be implemented soon?
I'm currently having the same issue as @Brammm and having just and "include" param would be perfect.

mrtus:
👋 We will archive this repository and thus close this Pull Request.
You can read more about it in the [README](https://github.com/teamleadercrm/api/blob/master/README.md)

PR #536, opened 2020-06-08
Allow filtering contacts and companies on segment_id

Brammm:
Also make segments discoverable, this is V1 functionality that is not yet available in V2.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1009,6 +1009,34 @@ Get a list of all work types, sorted alphabetically (on their name).
                 + id: `811a5825-96f4-4318-83c3-2840935c6003` (string)
                 + name: `Planning` (string)
 
+## Segments [/segments]
+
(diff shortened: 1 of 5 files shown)


mrtus (commented):
Hi @Brammm!

Thank you for thinking with us. 

I am afraid we won't be implementing functionality regarding segments any time soon, as currently, we do not have a clear idea yet where we want to go with the concept of segments. Looks like we already linked you to this as part of a feature request, so if we do decide to implement this we should be given an update.

At the moment we are aiming to use filters on `.list` API endpoints to create this filtering behaviour, a great example of how we use these filters is visible on our new deal-pipeline.

Brammm:
That's annoying to hear. Our customer will be using segments heavily in the CRM, something we need to bring to our app as well (which we are building an integration for).

PR #507, opened 2020-03-09
IN-2616 Add link type deal to events.list

luisaterrabrito:

PR #537, opened 2020-06-10
In-2616 allow filtering based on a linked deal on events list

EOSullivanBerlin:
This is need for the new deal detail page. 
We want to show our customers the events that are linked to that deal.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,7 +374,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
-- We added filtering on a linked `deal` to `events.list`.
+- We added filtering on a new `link` type `deal` to `events.list`.
 
 #### May 2020
 
@@ -2166,6 +2166,7 @@ Get a list of calendar events.
                             + Members
                                 + company
                                 + contact
+                                + deal
                         + id: `c9258836-f9a5-40cb-aa2a-d55c22991b93` (string)
 
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -372,6 +372,10 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+#### June 2020
+
+- We added filtering on a linked `deal` to `events.list`.
+
 #### May 2020
 
 - We added the `purchase_order_number` field to `projects.info`, `projects.create`, and `projects.update`
@@ -2111,6 +2115,7 @@ Get a list of calendar events.
                     + Members
                         + contact
                         + company
+                        + deal
                 + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
             + task_id: `5309c8c9-e313-47b0-90ba-6850c3dc3e33` (string, optional)
         + page (Page, optional)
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -372,6 +372,10 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+#### June 2020
```
luisaterrabrito: I think we call it Latest usually

lorgan3 (dismissed):

apiary.apib
EOSullivanBerlin: Okay I will change it to latest.

lorgan3 (dismissed):

lowiebenoot (dismissed):

src/04-calendar/events.apib
```diff
@@ -28,6 +28,7 @@ Get a list of calendar events.
                     + Members
                         + contact
                         + company
+                        + deal
```
dietrichm, line 31: This exposes that we will start having `deal` type links, so we can add that entity type to the enum of `links` types on lines 78 and 122.

src/changes-backwards-compatible.apib
```diff
@@ -4,6 +4,10 @@
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+#### Latest
+
+- We added filtering on a linked `deal` to `events.list`.
```
dietrichm: PP: let's make it explicit that the events endpoints now have support for deal links, so we added `link` as a link type plus allow to filter on it.

dietrichm:
There's some overlap with https://github.com/teamleadercrm/api/pull/507. I propose to merge that one and rebase this PR.

EOSullivanBerlin:
@dietrichm You are right the branches are very similar.

lowiebenoot (approved):

lorgan3 (approved):

lowiebenoot (approved):

src/04-calendar/events.apib
dietrichm: What is the reason that we only have the new `deal` link type on the list endpoint and not on the info one?

src/changes-backwards-compatible.apib
```diff
@@ -4,6 +4,10 @@
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+#### Latest
+
+- We added filtering on a new `link` type `deal` to `events.list`.
```
dietrichm, line 9: Filtering is not the main takeaway I think. Might be better to word this as:
```suggestion
- As events can be linked to deals, we added the `deal` link type to `events.list` and `events.info`, and allow filtering on deals in `events.list`.
```

dietrichm (commented):
As #507 got merged, can we rebase this PR to master and fix the missing deal link type for the info endpoint?

src/04-calendar/events.apib
EOSullivanBerlin: We are only using it on the list. Does it make sense to add it to an end point if we are not using it. I assume this is a case of YAGNI You aren't going to need it.

src/changes-backwards-compatible.apib
EOSullivanBerlin: I will use your description but not mention the info as we have not added that yet.

src/changes-backwards-compatible.apib
EOSullivanBerlin: Okay

PR #538, opened 2020-06-11
scheduled bool filter on tasks.list

EOSullivanBerlin:
Allow the API user to filter based on if the tasks are scheduled or not.

### DIFF
--- a/src/08-tasks/tasks.apib
+++ b/src/08-tasks/tasks.apib
@@ -12,6 +12,7 @@ Get a list of tasks.
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional, nullable) - Returns tasks that are assigned to this user or to a team to which this user belongs. When passing `null`, it returns tasks that are unassigned.
             + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
             + completed: true (boolean, optional)
+            + scheduled: true (boolean, optional)
             + due_by: `2019-02-04` (string, optional)
             + due_from: `2019-10-20` (string, optional)
             + term: `website design` (string, optional) - Searches for a term in the description.
--- a/apiary.apib
+++ b/apiary.apib
@@ -374,6 +374,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 #### Latest
 
+- We added a filter `scheduled` to `tasks.list`.
+
 - We added filtering on a linked `deal` to `events.list`.
 
 #### May 2020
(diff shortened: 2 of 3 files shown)


lorgan3 (commented):
The `apiary.apib` should also be updated right?

luisaterrabrito (approved):

src/08-tasks/tasks.apib
```diff
@@ -12,6 +12,7 @@ Get a list of tasks.
             + user_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional, nullable) - Returns tasks that are assigned to this user or to a team to which this user belongs. When passing `null`, it returns tasks that are unassigned.
             + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
             + completed: true (boolean, optional)
+            + scheduled: true (boolean, optional)
```
dietrichm, line 15: Q: should we make the value of `scheduled` explicit in the task data in the info and list endpoint?

dietrichm (changes requested):
Main Apiary file was not recompiled yet, indeed.

src/08-tasks/tasks.apib
EOSullivanBerlin: Are you suggesting that the info and list requests would return something like: 
`scheduled: true`?

adognameddog: > Are you suggesting that the info and list requests would return something like:
> `scheduled: true`?

Indeed. I would also go that route if the implementation isn't too complicated.

dietrichm: PP: have the `scheduled` property already in the list and info endpoint return data. This will also help us with testing.

dietrichm (approved):

lorgan3 (approved):

adognameddog (approved):

src/08-tasks/tasks.apib
mrtus, line 15: Q: Can completed & scheduled occur at the same time? If not should we make this a "status" property instead?

lowiebenoot: Yes, you can complete a task without scheduling it.

EOSullivanBerlin: It's actually quite a lot of work to return the scheduled property. It can be done for sure but it involves joining different tables. Which I find a bit difficult. 
I can do it but I think I will have to ask Pieter if this has priority for my time. 

Due to merge conflicts on this branch I will close this Pr and open a new one.

PR #540, opened 2020-06-18
Set a project's end date

ketlane:
This PR adds the ability to set a project's end date on `projects.create` and `projects.update`. The field is already exposed in `projects.list` and `projects.info` because we calculate it behind the scenes as the last milestone's due date.

We're working on allowing to explicitly set the field.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3250,7 +3250,7 @@ Create a new project.
         + title: `New company website` (string, required)
         + description (string, optional)
         + starts_on: `2020-06-15` (string, required)
-        + due_on: `2020-09-15` (string, optional)
+        + due_on: `2020-09-15` (string, optional) - Defaults to the last milestone's due date when omitted
         + milestones (array, required) - At least one milestone is required
             + (object)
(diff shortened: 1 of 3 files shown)


joaoinacio (approved):

mikeverf (approved):

ArnaudWeyts (approved):

Pegasuz (approved):

apiary.apib
```diff
@@ -3247,7 +3249,8 @@ Create a new project.
     + Attributes (object)
         + title: `New company website` (string, required)
         + description (string, optional)
-        + starts_on: `2016-02-04` (string, required)
+        + starts_on: `2020-06-15` (string, required)
+        + due_on: `2020-09-15` (string, optional)
```
adognameddog: Q:  should we mention that the `due_on` date will default to the last phase's due date it this value is omitted?

ketlane: Ok, I documented this.

adognameddog (approved):

PR #541, opened 2020-06-18
Archived departments

mrtus:
### Added
- A status property to which shows the status of that department

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -375,7 +375,7 @@ We list all backwards-compatible additions here. These are currently available i
 ### Latest
 
 - We added the `emails` property to `departments.list`.
-- We added the `status` property to `departments.info` and `departments.list`, this property is also available as a filter on `departments.list`. Only active departments are shown by default in `departments.list`.
+- We added the `status` property to `departments.info` and `departments.list`, this property is also available as a filter on `departments.list`.
 
 #### May 2020
 
--- a/apiary.apib
+++ b/apiary.apib
@@ -705,6 +705,10 @@ Get a list of departments.
                                 + primary
                                 + invoicing
                         + email: `info@piedpiper.eu` (string)
+                + status: `active` (enum)
+                    + Members
+                        + active
+                        + archived
 
 ### departments.info [GET /departments.info]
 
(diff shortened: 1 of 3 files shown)


nielsvermaut (dismissed):

apiary.apib
```diff
@@ -676,6 +677,10 @@ Get a list of departments.
     + Attributes (object)
         + filter (object)
             + ids: `92296ad0-2d61-4179-b174-9f354ca2157f`,`53635682-c382-4fbf-9fd9-9506ca4fbcdd` (array[string], optional)
+            + status (array[enum], optional) - Filters on status
```
rainbow-alex, line 680: It should be specified what the default behavior of this filter is (`active` if I understand correctly)

There should be (as I believe @mrnnn mentioned) a way to query active+archived together.

rainbow-alex (changes requested):

apiary.apib
mrtus: I'm not sure what you mean with default behaviour of a filter, it's either filled in or not. Did you maybe mean the default behaviour of the .list endpoint itself?


Since this property is an array the behaviour of querying active & archived together is already available.

rainbow-alex: If it's not filled in the default is `[active]` not `[active, archived]` as you'd expect from this specification.

mrtus: Right, I didn't think about the empty array possibility. 👍

mrtus: I think we don't need to describe this behaviour though, looking at the other other endpoints we never describe any behaviour, as I slightly seem to remember this documentation is about possible request & response properties and not about documenting behaviour on/with those properties to avoid this becoming a knowledge base of features.

Although do you have a proposal how we could maybe phrase this?

tiagobastos (dismissed):

steftrenson (dismissed):

apiary.apib
rainbow-alex: Something like:

```
(default: `[active]`)
```

would be plenty I think?

mrtus: This remark became obsolete after always returning active & archived entities by default

tiagobastos (approved):

rainbow-alex (approved):

steftrenson (approved):

PR #542, opened 2020-06-22
In-2616 inform the users that they can filter on events list  by linked deals

EOSullivanBerlin:
Based on this closed branch: https://github.com/teamleadercrm/api/pull/537

### DIFF
--- a/src/changes-backwards-compatible.apib
+++ b/src/changes-backwards-compatible.apib
@@ -6,7 +6,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
-- As events can be linked to deals, we added the `deal` link type to `events.list` and allow filtering on deals in `events.list`.
+- As events can be linked to deals, we added the `deal` link type to `events.list` and `events.info`, and allow filtering on deals in `events.list`.
 
 - We added the `emails` property to `departments.list`.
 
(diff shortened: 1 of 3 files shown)


lorgan3 (commented):
Shouldn't you also add the `deal` type to the events `link` field? I don't fully understand all the discussions in your other pr 😅

EOSullivanBerlin:
@lorgan3 you're right its being returned on the info too.

lorgan3 (dismissed):

lowiebenoot (dismissed):

src/changes-backwards-compatible.apib
```diff
@@ -6,6 +6,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
+- As events can be linked to deals, we added the `deal` link type to `events.list` and `events.info`, and allow filtering on deals in `events.list`.
```
dietrichm: The new `deal` link type and filter was added in #507 and hence is mentioned in the changes for February 2020. We should only adjust that change's description and indeed add the `deal` link type for the info endpoint.

So, I think we should rather reword the change in February instead of listing this as a new change that is yet to be released (which is not true).

luisaterrabrito (approved):

lorgan3 (approved):

lowiebenoot (approved):

dietrichm (approved):
👍

PR #547, opened 2020-07-02
[IN-476] Add quotations.update blueprint

lorgan3:


luisaterrabrito (dismissed):

lowiebenoot (dismissed):
`title` will just be added later, right? When we actuall have titles.

lorgan3:
@lowiebenoot Yes, I assume this will be part of the tickets when we're actually allowing the title to be changed.

EOSullivanBerlin (dismissed):

src/03-deals/quotations.apib
```diff
@@ -156,4 +156,41 @@ Download a quotation in a specific format.
...
+                                + Members
+                                    + percentage - Values between 0 and 100
+        + discounts (array, optional)
+            + (object)
+                + value: 10 (number, required)
+                + type (enum, required)
+                    + Members
+                        + percentage - Values between 0 and 100
+                + description: `winter promotion` (string, required)
+        + text: `Quotation text` (string, optional) - Uses Markdown formatting. A quotation needs `grouped_lines` and/or `text` to be valid
```
dietrichm: The text property should take `null` to allow it to be unset. (Not providing it will not update the property.)
```suggestion
        + text: `Quotation text` (string, optional, nullable) - Uses Markdown formatting. A quotation needs `grouped_lines` and/or `text` to be valid
```

src/03-deals/quotations.apib
```diff
@@ -156,4 +156,41 @@ Download a quotation in a specific format.
...
\ No newline at end of file
+            + id: `e4bf74ca-e900-471f-84b1-276e5d3afae4` (string)
+
+### quotations.update [POST /quotations.update]
+
++ Request (application/json)
+
+    + Attributes (object)
+        + quotation_id: `5b16f6ee-e302-0079-901b-50c26c4a55b1` (string, required)
+        + `grouped_lines` (array, optional)
```
dietrichm: I'd keep the remark here that either grouped lines or text is required for a quotation.
```suggestion
        + `grouped_lines` (array, optional) - A quotation needs `grouped_lines` and/or `text` to be valid
```

src/changes-backwards-compatible.apib
```diff
@@ -10,6 +10,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 - We added `quotations.create`.
 
+- We added `quotations.update`.
```
dietrichm: Can we summarise these three changes into one? They're all about adding quotations endpoints.

dietrichm (changes requested):

src/03-deals/quotations.apib
lorgan3: I suppose you can update the discounts without touching the line items or text so I don't think so?

src/03-deals/quotations.apib
lorgan3: Allowing this will require extra validation so you don't end up without product lines and no text, should we do that for the first version?

dietrichm: The business rule of having either grouped lines or text or both indeed still needs to stand, and we should also not disallow to remove a text from a quotation again through the update endpoint. The latter would be pretty weird for an integrator.

So I think this is some essential behaviour to publicly release the endpoint.

(Basically we already need similar validation rules as passing an empty string for a text would create the same behaviour as passing `null`.)

src/03-deals/quotations.apib
dietrichm: Indeed that should be possible. So the question is: does this remark refer to the (updated) quotation or only to the provided data? I assume it refers to the quotation.

I'm OK with removing it provided we also clear it here: https://github.com/teamleadercrm/api/pull/547/files#diff-2625016b50d68d922257f74801cac29cR2240

src/03-deals/quotations.apib
```diff
@@ -156,4 +156,41 @@ Download a quotation in a specific format.
...
-            + id: `e4bf74ca-e900-471f-84b1-276e5d3afae4` (string)
\ No newline at end of file
+            + id: `e4bf74ca-e900-471f-84b1-276e5d3afae4` (string)
+
+### quotations.update [POST /quotations.update]
+
++ Request (application/json)
+
+    + Attributes (object)
+        + quotation_id: `5b16f6ee-e302-0079-901b-50c26c4a55b1` (string, required)
```
dietrichm: Didn't notice it in the first review, but the subject's ID is usually passed as `id`.
```suggestion
        + id: `5b16f6ee-e302-0079-901b-50c26c4a55b1` (string, required)
```

dietrichm (changes requested):

src/03-deals/quotations.apib
lorgan3: I'll put it back if you think it's clearer that way

luisaterrabrito (approved):

dietrichm (approved):

lowiebenoot (approved):

PR #557, opened 2020-09-09
Add projects.close endpoint

joaoinacio:


ArnaudWeyts (dismissed):

mikeverf (dismissed):

src/07-projects/projects.apib
```diff
@@ -211,6 +211,18 @@ Update a project.
 + Response 204 (application/json)
 
 
+### projects.close [POST /projects.close]
+
+Closes a project, and all its phases
```
adognameddog: I think we should also mention it closes tasks, but not meetings.

adognameddog (dismissed):

mikeverf (approved):

apiary.apib
```diff
@@ -3486,6 +3487,18 @@ Update a project.
...
+### projects.close [POST /projects.close]
+
+Closes a project, all its phases, and all tasks within each phase (but not meetings).
+
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `01548b10-4932-4a18-8a89-005ad09db2c8` (string, required)
+
++ Response 204 (application/json)
```
tiagobastos, line 3499: `204 - No content` has no content so the `application/json` response doesn't apply right?

joaoinacio: > `204 - No content` has no content so the `application/json` response doesn't apply right?

Good point, I think you may be right (the content-type shouldn't be on the response).
OTOH I copy-pasted it, so would probably make sense to do a follow-up PR fixing this on every endpoint.

tiagobastos: Good with a follow-up.

tiagobastos (approved):

adognameddog (approved):
:white_check_mark: Manually checked

ArnaudWeyts (approved):

PR #562, opened 2020-09-17
[INS-109] Add purchase_price to quotations.create/update

rainbow-alex:
Jira: https://teamleader.atlassian.net/browse/INS-109

implementation: https://github.com/teamleadercrm/core/pull/17554

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -375,7 +375,7 @@ We list all backwards-compatible additions here. These are currently available i
 ### Latest
 
 - We added the `discounts` field to `quotations.info`.
-- We added the `purchase_price` field to `quotations.list` and `quotations.info`.
+- We added the `purchase_price` field to `quotations.list`, `quotations.info`, `quotations.create` and `quotations.update`.
 - We added `projects.close` and `projects.reopen` endpoints.
 
 ### August 2020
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -2217,6 +2217,7 @@ Create a quotation.
                                 + Members
                                     + percentage - Values between 0 and 100
                         + product_id: `d905ff57-e866-0f59-9d1e-1fd4538bfae1` (string, optional) - This is purely informational and does not affect the quotation besides adding a reference to the product
+                        + purchase_price (Money, nullable, optional) - the currency must match the account currency
```
lorgan3, line 2220: PP: there's no point in being able to pass null when creating

lorgan3 (commented):
Will you also make it so that if the items are in a different currency than your account, adding a purchase price is not supported? This is a change we recently made in the ui because that case doesn't make sense in the quotation editor.

Also don't forget to update the changelog!

apiary.apib
rainbow-alex: Seems nice to me that (as a client) you don't have to remove the key-value pair if the value happens to be null.

luisaterrabrito (approved):

lorgan3 (approved):

EOSullivanBerlin (approved):

steftrenson (approved):

jnstr (approved):

PR #569, opened 2020-10-10
Customer on projects.update [PRO-4644]

adognameddog:


src/07-projects/projects.apib
```diff
@@ -205,6 +205,12 @@ Update a project.
                 + done
                 + cancelled
         + starts_on: `2016-02-04` (string)
+        + customer (object) - Null will unlink the customer from the project
```
mrtus: Q: Reading the extra information given, did you mean to make customer nullable?
```apib
        + customer (object, nullable) - Null will unlink the customer from the project
```

adognameddog: I did that, but then I saw this does nothing in the generated documentation, it does not mention it is nullable 🤷

joaoinacio (approved):

joaoinacio:
✅ Manually checked.

src/07-projects/projects.apib
mrtus: That is weird indeed 🤔 I thought this was visible somewhere, it should be at least in the Json Schema itself.

mikeverf (approved):

ArnaudWeyts (approved):

PR #570, opened 2020-10-10
Budget on projects.update [PRO-4646]

adognameddog:


src/07-projects/projects.apib
```diff
@@ -211,6 +211,7 @@ Update a project.
                     + contact
                     + company
             + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
+        + budget (Money) - Null will reset the budget to zero
```
ArnaudWeyts: Should we really allow passing null trough the API, since it's not technically nullable? I would always require a positive number or 0 instead, until it's possible to make the budget truly nullable, that way we don't create a "breaking change" if we do decided to allow null

mikeverf (dismissed):

src/07-projects/projects.apib
adognameddog: Good point.

joaoinacio (dismissed):

ArnaudWeyts (approved):

joaoinacio (approved):

mikeverf (approved):

PR #580, opened 2020-10-30
[INS-46] proposal for adding the related project to the invoices list end point

EOSullivanBerlin:
Ignore the commit message its on the `invoice.list` where I need it.


rainbow-alex (changes requested):
Not needed, it's implemented for internal use, behind an include. reasoning: it might still become a list.

EOSullivanBerlin:
@rainbow-alex yeah sorry.

PR #582, opened 2020-11-03
Add filter for active companies and contacts

Brammm:
Teamleader's import and interface allows you to deactivate companies and contacts, which is also visually shown very well in the interface:

However, there is no way to separate these inactive companies/contacts in the API, unless you decide to make a segment for them and use API v1...

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1108,6 +1108,7 @@ Get a list of contacts.
             + term: `James` (string, optional) - Filters on first_name, last_name, email and telephone
             + updated_since: `2016-02-04T16:44:33+00:00` (string, optional)
             + tags: `expo`,`prospect` (array[string], optional) - Filters on contacts coupled to all given tags
+            + active: `true` (boolean, optional) - Filters on a contact's active status
         + page (Page, optional)
         + sort (array, optional)
             + (object)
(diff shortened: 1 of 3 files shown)


marievandaele:
Feel free to send an email to support@teamleader.eu, asking to be linked  to issue ID-5127 (add deactivate status to the contacts and companies endpoints). That way we'll update you once this has been added to our API

Brammm:
@marievandaele @mrtus I've just realised there isn't an `active` key in the `companies.info` endpoint, so you can't even see it in the API if a company was deactivated or not. 

This seems like an incredibly trivial thing to add, is there any chance of getting that?

mrtus:
> @marievandaele @mrtus I've just realised there isn't an `active` key in the `companies.info` endpoint, so you can't even see it in the API if a company was deactivated or not.
> 
> This seems like an incredibly trivial thing to add, is there any chance of getting that?

Hi @Brammm

It is best to direct your feature request through a support ticket, our support colleagues will then link your request to the correct issues.

Brammm:
Getting a big fat "get stuffed" from them...

MathiasDL:
@Brammm This has been added to the roadmap. You can ask TL support to link your customers to ATL-646 and ATL-647.

Note that the filter won't be called `active`, but `status` to make it consistent with the `status` field which is used now.

mrtus:
Hi @Brammm , this has been implemented here: https://github.com/teamleadercrm/api/pull/702

As a result I will close this change request

Brammm:
@mrtus Cool. We've since moved away from TL 😬

PR #585, opened 2020-11-19
[IN-3255] proposal for returning all user emails

lorgan3:
I kept the format like the profile endpoint but left out the data I didn't need.


rainbow-alex (approved):

src/01-general/users.apib
```diff
@@ -114,6 +124,11 @@ Get details for a single user.
             + first_name: `John` (string)
             + last_name: `Smith` (string)
             + email: `john@teamleader.eu` (string)
+            + emails (array)
+                + (object)
+                    + email: `john@teamleader.eu` (string)
```
luisaterrabrito, line 129: Would this replace the email? Should we state which one is the primary one?

lorgan3: It replaces `email` yes but we can't remove it because that would be a breaking change

> Should we state which one is the primary one?

The one in `email` is the primary one. I can also add it in the emails object but I didn't because it didn't seem super useful to me

lorgan3:
So in the end we will only allow sending from the primary email so this is useless now...
Sorry for the wasted review! 🙈

PR #600, opened 2021-08-03
Add is_duration_time_tracking to time tracking list

EOSullivanBerlin:
I need this value for the new time tracking overview.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -399,6 +399,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
+- We added `is_duration_time_tracking` on `timeTracking.list`
 - We added `term` to the filters on `workTypes.list`.
 - We added `started_on` to `timeTracking.update` and `timeTracking.add`.
 - We added the `quotations.delete` endpoint.
@@ -4238,6 +4239,7 @@ Get a list of tracked time.
                     + type: `workType` (string)
                     + id: `8a146eda-c9af-4b11-b63a-ddac47bb254d` (string)
                 + started_on: `2017-04-26` (string)
+                + is_duration_time_tracking: true (boolean)
                 + started_at: `2017-04-26T10:01:49+00:00` (string, nullable)
                 + ended_at: `2017-04-26T11:01:49+00:00` (string, nullable)
                 + duration: 3600 (number) - In seconds
--- a/apiary.apib
+++ b/apiary.apib
@@ -400,6 +400,8 @@ We list all backwards-compatible additions here. These are currently available i
 ### Latest
 
 - We added `is_duration_time_tracking` on `timeTracking.list`
+
+### June
 - We added `term` to the filters on `workTypes.list`.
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -399,6 +399,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
+- We added `is_duration_time_tracking` on `timeTracking.list`
```
lorgan3, line 402: I think the stuff below here probably needs a `### July` header

EOSullivanBerlin: Consider it done.

EOSullivanBerlin:
Q: do i need to add a test for this? None of the current test break when I run them and there doesn't seem to be an obvious place to check that it's being passed. The test mostly look at filters and not the contents of the timetracking.

lorgan3:
@EOSullivanBerlin For quotations we check all expected fields are present, you could do something similar here https://github.com/teamleadercrm/api-tests/blob/master/src/Scenarios/Quotations/ListTest.php#L52-L94

Also there's 1 more thing I thought about 😅 
You're adding a new field to the list endpoint but not info. It doesn't really make sense for the list to return more details than info so IMO you should add it to both endpoints

src/08-time-tracking/time-tracking.apib
```diff
@@ -66,6 +66,7 @@ Get a list of tracked time.
                     + type: `workType` (string)
                     + id: `8a146eda-c9af-4b11-b63a-ddac47bb254d` (string)
                 + started_on: `2017-04-26` (string)
+                + is_duration_time_tracking: true (boolean)
```
mrtus, line 69: If we add it to the `.list` endpoint let's also add this to the `.info` endpoint.

mrtus: Not sure a boolean is the best property to use here though. What about a type/...? Which is "duration" and not sure what the alternative would be?

mrtus: Not sure a boolean is the best property to use here though. What about "type" / or something that reflects this more? Where the value is "duration" and not sure what the alternative would be, how do we call the "normal" variant?

There could maybe be more time-tracking types in the future. We don't want exclusive true boolean values for those property then.

ArnaudWeyts: The other type is start-stop, currently our application is built around these 2 types, so I don't think another type will be added anytime soon

ArnaudWeyts:
I'm not sure if we should already add this publicly, maybe we should only provide this value for internal use first?

src/08-time-tracking/time-tracking.apib
ArnaudWeyts: After discussing with @mrtus, let's go for a `type` property with either `start_stop` (?) or `duration`, and only return this value internally for now

EOSullivanBerlin: I will make it a type.

PR #608, opened 2021-11-04
[CORE-3693] Add 'status' property to companies list info

sanderbrugge:


apiary.apib
```diff
@@ -1519,6 +1521,7 @@ Get a list of companies.
             + (object)
                 + id: `96a38bbf-24ed-4083-8a5c-20db92aa471e` (string)
                 + name: `Pied Piper` (string)
+                + status: `active` (string)
```
tiagobastos: Let's be explicit and list all the variations that `status` can take:            
```suggestion
                + status: `active` (enum[string])
                    + Members
                        + `active`
                        + ...
```

apiary.apib
```diff
@@ -399,6 +399,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
+- We added `status` on `companies.list`
+- We added `status` on `companies.info`
```
mrtus, line 403: Q: Let's maybe create a new header for this and wrap the older changes under their month

mrtus: Let's apply this change in the changelog file and not in the "build" file here though.

mrtus (commented):
👍

apiary.apib
JordyDeGroeve: Oh missed that.

p-ba (approved):

tiagobastos (approved):
✅

JordyDeGroeve (approved):

PR #615, opened 2021-11-29
[ATL-547] Discounts on creditNotes.info

MathiasDL:
Best reviewed commit per commit since I packed some cleanups into this PR.

### DIFF
--- a/src/05-invoicing/credit-notes.apib
+++ b/src/05-invoicing/credit-notes.apib
@@ -104,6 +104,13 @@ Get details for a single credit note
                         + id: `dc2a40bf-f050-4c8e-a002-872a051150b9` (string)
                     + email: `duivels@test.com` (string, nullable)
                     + national_identification_number: `123` (string, nullable)
+                + discounts (array, optional)
+                    + (object)
+                        + value: 10 (number, required)
+                        + type (enum, required)
+                            + Members
+                                + percentage - Values between 0 and 100
+                        + description: `winter promotion` (string, required)
                 + total (object)
                     + tax_exclusive (Money)
                     + tax_inclusive (Money)
--- a/src/changes-backwards-compatible.apib
+++ b/src/changes-backwards-compatible.apib
@@ -4,7 +4,7 @@
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
-### Latest
+### November 2021
 
 - We added `status` on `companies.list`.
 - We added `status` on `companies.info`.
--- a/src/changes-backwards-compatible.apib
+++ b/src/changes-backwards-compatible.apib
@@ -6,12 +6,12 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### Latest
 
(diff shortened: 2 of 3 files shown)


mrtus:
> Best reviewed commit per commit since I packed some cleanups into this PR.

Let's maybe split that up next time? 😉 If we now want to unpublish this again we lose the adjustments as well.

src/changes-backwards-compatible.apib
```diff
@@ -6,12 +6,13 @@ We list all backwards-compatible additions here. These are currently available i
...
-- We added `status` on `contacts.info`
-- We added `ids` to the filters on `creditNotes.list`
-- We added `invoice_id` and `project_id` to the filters on `creditNotes.list`
+- We added `status` on `companies.list`.
+- We added `status` on `companies.info`.
+- We added `status` on `contacts.list`.
+- We added `status` on `contacts.info`.
+- We added `ids` to the filters on `creditNotes.list`.
+- We added `invoice_id` and `project_id` to the filters on `creditNotes.list`.
+- We added `discounts` to `creditNotes.info`.
```
mrtus, line 15: Q: Should we add this to a December header already?

ArnaudWeyts (dismissed):

src/changes-backwards-compatible.apib
MathiasDL: No, since December is 2 days away and the code has already been deployed

MathiasDL:
> > Best reviewed commit per commit since I packed some cleanups into this PR.
> 
> Let's maybe split that up next time? 😉 If we now want to unpublish this again we lose the adjustments as well.

Sorry, didn't feel like merge conflicts today 😬

src/changes-backwards-compatible.apib
mrtus: Then a November header maybe?

MathiasDL: 🤦  I'll dig out which ones were from November and update

MathiasDL: There were no updates no October, all are from November. I changed the _Latest_ segment to _November 2021_.

This seems like an easier way to work instead of doing a separate PR at the start of every month: just don't use a Latest header, always use the header of the current month. The first creator of a PR in a new month (and the reviewers) should notice that the current month is not in there yet, and thus create a new section for the current month.

src/05-invoicing/credit-notes.apib
```diff
@@ -104,6 +104,13 @@ Get details for a single credit note
                         + id: `dc2a40bf-f050-4c8e-a002-872a051150b9` (string)
                     + email: `duivels@test.com` (string, nullable)
                     + national_identification_number: `123` (string, nullable)
+                + discounts (array, optional)
+                    + (object)
```
joaoinacio, line 108: Nitpick: this could/should be a reusable object

joaoinacio (approved):

src/05-invoicing/credit-notes.apib
MathiasDL: Ok, I'll do it in a followup PR as this one is already loaded with some extra stuff

mrtus (approved):

sanderbrugge (approved):

PR #625, opened 2022-02-15
[ATL-608] Moving project from Internal API to API 

gdegirmenci:


Jdww (dismissed):

.gitignore
```diff
@@ -1,2 +1,3 @@
 /node_modules
 apiary.html
+.idea
```
tiagobastos: Don't think this belongs here though. 
Different editors will create similar folders, meaning we would need to keep updating this file to account for external tools. 

Add this to your `.gitignore_global` and keep this file specific to the repo internal usage.

tiagobastos (commented):
You missed adding this to the changelog.

tiagobastos (approved):

Jdww (approved):

JordyDeGroeve (approved):

p-ba (approved):

PR #626, opened 2022-03-03
Document subscriptions.list

JordyDeGroeve:


rainbow-alex (approved):

gdegirmenci (approved):

src/05-invoicing/subscriptions.apib
```diff
@@ -0,0 +1,74 @@
...
+                    + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
+                + total (object)
+                    + tax_exclusive (Money)
+                    + tax_inclusive (Money)
+                    + taxes (array)
+                        + (object)
+                            + rate: 0.21 (number)
+                            + taxable (Money)
+                            + tax (Money)
+                + `web_url`: `https://app.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
```
joaoinacio, line 73:
```suggestion
                + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
```

apiary.apib
```diff
@@ -3289,6 +3289,81 @@ Download a credit note in a specific format.
...
++ Response 200 (application/json)
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `e2314517-3cab-4aa9-8471-450e73449041` (string)
+                + department (object)
+                    + type: `department` (string)
+                    + id: `5e90eb0a-b502-4344-aa0f-3b8525af6186` (string)
+                + title: `Subscription for cookies` (string)
+                + active: true (boolean)
```
mrtus, line 3322: In the future there could be a feature introduced to temporarly suspend subscriptions.

Let's change this `active` property to a `status` property instead?
```
                + status: active (enum)
```

It does not make sense to use a boolean in this case.

apiary.apib
```diff
@@ -3289,6 +3289,81 @@ Download a credit note in a specific format.
...
+                + next_renewal_date: `2020-06-21` (string)
+                + invoicee (object)
```
mrtus, line 3324: Q: I don't think `invoicee` is the correct term here since there is no invoice, what about `subscriber` or `invoice_recipient` could be one? Can we think of a better name maybe?

mrtus (changes requested):
Let's make sure we get that `active` boolean property replaced with an enum value.

apiary.apib
JordyDeGroeve: 100% correct. Will update the usages and the internal structure first. 👍

apiary.apib
JordyDeGroeve: `invoice recipient` is literally the description when you look up the term `invoicee`.
To me, this does not feel like an improvement.

`Subscriber`, on the other side, is more into context. Let's check with @Miljar on this.

rainbow-alex: It's the invoicee on the invoices that are generated, seems fine to me.

mrtus: > `invoice recipient` is literally the description when you look up the term `invoicee`.

Ya, I realised that as well yesterday after closing my laptop 😂.

JordyDeGroeve: @tiagobastos also noticed it is in line with the object (containing a **customer and FAO**) that is returned in `invoices.list`

JordyDeGroeve:
Going to close this (for now) and realign with the team on how to polish the internal endpoint before publishing it

PR #627, opened 2022-03-09
Document the subscription endpoints

JordyDeGroeve:
### Add documentation for the following endpoints:
- `subscriptions.list`
- `subscriptions.info`

This PR also adds the Customer to the datastructures.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -3289,6 +3289,144 @@ Download a credit note in a specific format.
             + location: `https://cdn.teamleader.eu/file` (string) - A temporary url where the requested file can be downloaded
             + expires: `2018-02-05T16:44:33+00:00` (string) - Expiration time of the temporary download link
 
+## Subscriptions [/subscriptions]
+
+### subscriptions.list [GET /subscriptions.list]
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + ids: `46156648-87c6-478d-8aa7-1dc3a00dacab`, `46156648-87c6-478d-8aa7-1dc3a00daca4` (array[string], optional)
+            + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Filter on subscriptions created from an invoice
+            + deal_id: `fddf42f9-bda1-0c49-a22b-0bf147cd1e3f` (string, optional) - Filter on subscriptions created from a deal
+            + customer (Customer)
+        + page (Page, optional)
+        + sort (array, optional)
+            + (object)
+                + field (enum, required)
+                    + Members
+                        + title
+                        + created_at
+                + order (enum, optional)
+                    + Members
+                        + asc
+                        + desc
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `e2314517-3cab-4aa9-8471-450e73449041` (string)
+                + title: `Subscription for cookies` (string)
+                + note: `'Some extra remarks about the subscription'` (string, nullable) - plaintext
+                + state: `active` (enum[string])
+                    + Members
+                        + active
+                        + deactivated
+                + department (object)
+                    + type: `department` (string)
+                    + id: `5e90eb0a-b502-4344-aa0f-3b8525af6186` (string)
+                + invoicee (object)
+                    + customer (Customer)
+                    + for_attention_of (object, nullable)
+                        + contact (object, nullable)
+                            + type: `contact` (string)
+                            + id: `d4391f46-a32c-458c-b2ee-833fd27a348d` (string)
+                        + name: `Radja Nainggolan` (string, nullable)
+                + project (object, nullable)
+                    + type: `project` (string)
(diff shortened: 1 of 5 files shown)


apiary.apib
```diff
@@ -3289,6 +3289,144 @@ Download a credit note in a specific format.
...
+                    + Members
+                        + active
+                        + deactivated
+                + department (object)
+                    + type: `department` (string)
+                    + id: `5e90eb0a-b502-4344-aa0f-3b8525af6186` (string)
+                + invoicee (object)
+                    + customer (Customer)
+                    + for_attention_of (object, nullable)
+                        + contact (object, nullable)
```
Jdww, line 3336: Q: Is this nullable? Will it not always be set if for_attention_of is not null?

apiary.apib
```diff
@@ -3289,6 +3289,144 @@ Download a credit note in a specific format.
...
+                        + asc
+                        + desc
+
++ Response 200 (application/json)
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `e2314517-3cab-4aa9-8471-450e73449041` (string)
+                + title: `Subscription for cookies` (string)
+                + note: `'Some extra remarks about the subscription'` (string, nullable) - plaintext
```
Jdww: Q: Do we have any guidelines on when to return something in markdown?

apiary.apib
JordyDeGroeve: Yes the FAO has an optional contact.

apiary.apib
JordyDeGroeve: From what I could find, the notes on subscriptions don't accept markdown (yet).

joaoinacio: We actually return markdown on all long description fields AFAIK, so should be consistent

apiary.apib
```diff
@@ -3289,6 +3289,144 @@ Download a credit note in a specific format.
...
 
+## Subscriptions [/subscriptions]
+
+### subscriptions.list [GET /subscriptions.list]
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + ids: `46156648-87c6-478d-8aa7-1dc3a00dacab`, `46156648-87c6-478d-8aa7-1dc3a00daca4` (array[string], optional)
+            + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Filter on subscriptions created from an invoice
```
rainbow-alex: the subscriptions aren't created by the invoice, it's the other way around

```suggestion
            + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Find subscriptions that generated the given invoice
```

apiary.apib
```diff
@@ -3289,6 +3289,144 @@ Download a credit note in a specific format.
...
+            + deal_id: `fddf42f9-bda1-0c49-a22b-0bf147cd1e3f` (string, optional) - Filter on subscriptions created from a deal
```
rainbow-alex, line 3305: Probably something similar here except I don't even know how the deal relates.

rainbow-alex (dismissed):

apiary.apib
JordyDeGroeve: Not sure this is correct. It is possible [to create a subscription from deal/invoice](https://support.focus.teamleader.eu/en/support/solutions/articles/7000017939-how-to-creating-a-subscription-in-teamleader-focus).
~We persist the **parent** invoice in this case, not the resulting invoices of that subscription.~
This filter does look at all invoices apparently..

apiary.apib
JordyDeGroeve: see https://github.com/teamleadercrm/api/pull/627#discussion_r823438962

apiary.apib
JordyDeGroeve: Let's update the naming to make it clearer that this is indeed filtering on related invoices (and not the actual parent invoice)

gdegirmenci (commented):
waiting changes

gdegirmenci (approved):

tiagobastos (approved):

mrtus (approved):

apiary.apib
```diff
@@ -3289,6 +3293,144 @@ Download a credit note in a specific format.
...
+## Subscriptions [/subscriptions]
+
+### subscriptions.list [GET /subscriptions.list]
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + ids: `46156648-87c6-478d-8aa7-1dc3a00dacab`, `46156648-87c6-478d-8aa7-1dc3a00daca4` (array[string], optional)
+            + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Find subscriptions that generated the given invoice
+            + deal_id: `fddf42f9-bda1-0c49-a22b-0bf147cd1e3f` (string, optional) - Filter on subscriptions created from a deal
```
mrtus, line 3305: Stijn and I once discussed we should maybe start making these about `ids` because it would make more sense to allow filtering on multiple ids at once, which would reduce the amount of API calls necessary.

Let's keep as is for now though, this can be an addition later.

PR #629, opened 2022-04-11
`subscription.list` | Make filter on status explicit

tiagobastos:


gdegirmenci (dismissed):

apiary.apib
```diff
@@ -3304,6 +3304,10 @@ Download a credit note in a specific format.
             + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Find subscriptions that generated the given invoice
             + deal_id: `fddf42f9-bda1-0c49-a22b-0bf147cd1e3f` (string, optional) - Filter on subscriptions created from a deal
             + customer (Customer)
+            + status (array[enum], optional) - Filters on status
+                + Members
+                    + active - Filters on active subscriptions
+                    + deactivated - Filters on archived subscriptions
```
joaoinacio: 🤔 archived? deactivated? inactive?

mrtus: I don't think it is archived, deactivated sounds more correct 👌

gdegirmenci (approved):

joaoinacio (approved):

apiary.apib
```diff
@@ -3304,6 +3304,10 @@ Download a credit note in a specific format.
             + invoice_id: `60f03745-4289-4ae7-8c74-d797860bc887` (string, optional) - Find subscriptions that generated the given invoice
             + deal_id: `fddf42f9-bda1-0c49-a22b-0bf147cd1e3f` (string, optional) - Filter on subscriptions created from a deal
             + customer (Customer)
+            + status (array[enum], optional) - Filters on status
```
JordyDeGroeve, line 3307: Q: Do we really need the description? It does not add anything

JordyDeGroeve (approved):

apiary.apib
tiagobastos: For consistency with the previous filter's description, I would leave them be.

Jdww (approved):

tiagobastos:
While MC we noticed a small inconsistency in "status" _members_ in this and other endpoints. 
Properties with `array[enum]` are adding `null` to the list of allowed members while that is not correct.

Fixing outside of this PR.

rainbow-alex (approved):

PR #630, opened 2022-04-12
`subscriptions.update` | Add documentation

gdegirmenci:


rainbow-alex (commented):
There are more properties than this, right? The line items for one?

JordyDeGroeve (dismissed):
Indeed, we might need additional fields

gdegirmenci:
afaik we'll implement them for later, cc @joaoinacio

p-ba (dismissed):

apiary.apib
```diff
@@ -3431,6 +3431,42 @@ Download a credit note in a specific format.
...
+                + three_yearly
+                + four_yearly
+                + five_yearly
+                + six_yearly
+                + seven_yearly
+                + eight_yearly
+                + nine_yearly
+                + ten_yearly
+        + custom_fields (array[CustomFieldValue], optional)
+        + start_date: `2022-06-21` (string, optional)
```
joaoinacio, line 3466: PP: move it upwards (before payment_term), for legibility?

joaoinacio (approved):

apiary.apib
```diff
@@ -3431,6 +3431,42 @@ Download a credit note in a specific format.
...
+                + yearly
+                + two_yearly
+                + three_yearly
+                + four_yearly
+                + five_yearly
+                + six_yearly
+                + seven_yearly
+                + eight_yearly
+                + nine_yearly
+                + ten_yearly
```
mrtus, line 3464: Q: Should we leave these set of values out for now? These don't look very eligable.

apiary.apib
```diff
@@ -3431,6 +3431,42 @@ Download a credit note in a specific format.
...
+    + Attributes (object)
+        + id: `e2314517-3cab-4aa9-8471-450e73449041` (string, required)
+        + title: `Subscription for cookies` (string, optional)
+        + project_id: `ebafa4c5-fa8a-4409-92e5-1b192243372f` (string, optional, nullable)
+        + payment_term (PaymentTerm, optional)
+        + periodicity: `monthly` (enum, optional)
+            + Members
+                + weekly
+                + two_weekly
+                + four_weekly
```
mrtus, line 3449: Q: What about a different structure? I find these enums a bit weird and limiting.

```
"periodicity": {
  "unit": "week",
  "period": 5
}
```

Using a different structure would allow more granular periodicities and would allow periodicities which are currently not available and might be fitting our customers needs better.

JordyDeGroeve (approved):
Agree with the suggested format of @mrtus

PR #631, opened 2022-04-21
Document `/contacts.updateCompanyLink` contract

tiagobastos:


joaoinacio (approved):

apiary.apib
```diff
@@ -1505,6 +1509,20 @@ Unlink a contact from a company.
...
 
+### contacts.updateCompanyLink [POST /contacts.updateCompanyLink]
+
+Update contact to company link.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `75596038-b9c8-4a37-969d-61059e300a28` (string, required)
+        + company_id: `4da20f00-0628-4336-b885-aa90e580dd85` (string, required)
```
mrtus, line 1520: Q: Should we clarify a bit more that this is the contact id?

apiary.apib
```diff
@@ -1505,6 +1509,20 @@ Unlink a contact from a company.
 
 + Response 204
 
+### contacts.updateCompanyLink [POST /contacts.updateCompanyLink]
```
mrtus, line 1512: Not sure about the naming but can't really think about any other?

Q: Thinking out loud, but maybe we could allow `.linkToCompany` to do an upsert to facilitate this instead?

What are our thoughts about this?

apiary.apib
```diff
@@ -1505,6 +1509,20 @@ Unlink a contact from a company.
...
+
+Update contact to company link.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `75596038-b9c8-4a37-969d-61059e300a28` (string, required)
```
mrtus, line 1519: Q: Should we clarify a bit more that this is the contact id?

mrtus (approved):

apiary.apib
mrtus: Not doing this anywhere else though so should be fine 👌

Jdww (approved):

apiary.apib
JordyDeGroeve: You can only link once imo. `SaveLinkToCompany` would do the trick, but is basically what we are doing here. (nothing against your proposal, just my 2 cents)

tiagobastos: It was considered during grooming to allow `.linkToCompany ` to become an upsert. 
Resoaning to dismiss it was that this would change the current endpoint behaviour. 

Up until now, any duplicated call to `.linkToCompany` would fail due to an already existing link.  
After this, it would persist the new link properties.

BrechtBonte (approved):

JordyDeGroeve (approved):

MathiasDL (approved):

PR #633, opened 2022-04-27
Document `/subscriptions.create` contract

gdegirmenci:


src/05-invoicing/subscriptions.apib
```diff
@@ -139,3 +139,33 @@
                         + taxable (Money)
                         + tax (Money)
             + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
+
+### subscriptions.create [POST /subscriptions.create]
+
```
tiagobastos, line 144: Let's add a small description of the purpose of the endpoint. 
It may seem redundant or implicit but it's the difference between a nice formatted UI. 
e.g.: 

| With Description  | Without Description |
| ------------- | ------------- |
| image  | image  |

PR #635, opened 2022-06-09
ATL-912 Document files endpoints

Jdww:


apiary.apib
```diff
@@ -4755,6 +4755,167 @@ Update the current timer. Only possible if there is a timer running.
...
+        + filter (object, required)
+            + subject (object, required)
+                + type (enum[string], required)
+                    + Members
+                        + company
+                        + contact
+                        + deal
+                        + invoice
+                        + creditNote
+                        + general_email_template
```
MathiasDL: B: enum format consistency: use CamelCase

Jdww: Good point, created a pr to change them all to camel case.

apiary.apib
```diff
@@ -4755,6 +4755,170 @@ Update the current timer. Only possible if there is a timer running.
...
+
+    + Attributes (object)
+        + id: `4afb0a9c-91c6-49ed-a2e5-ce7c1e3a87fb` (string, required)
+
++ Response 200 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + location: `https://cdn.teamleader.eu/file` (string) - A temporary url where the requested file can be downloaded
+            + expires: `2018-02-05T16:44:33+00:00` (string) - Expiration time of the temporary download url
```
mrtus: A more fitting name would be `expires_at` here. `expires` is more like a boolean.

```suggestion
            + expires_at: `2018-02-05T16:44:33+00:00` (string) - Expiration time of the temporary download url
```

apiary.apib
```diff
@@ -4755,6 +4755,170 @@ Update the current timer. Only possible if there is a timer running.
...
+        + subject (object, required)
+            + type: `company` (enum[string], required)
+                + Members
+                    + company
+                    + contact
+                    + deal
+                    + invoice
+                    + creditNote
+            + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+        + tag: `Mobile uploads` (string, optional)
```
mrtus: Q: Here we are calling it `tag`, but in `.info` & `.list` we call it `folder`? We should probably pick one?

apiary.apib
```diff
@@ -4755,6 +4755,170 @@ Update the current timer. Only possible if there is a timer running.
...
+
++ Response 200 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + location: `https://files.teamleader.eu/upload/66726166a` (string) - The temporary url where the proposed file can be uploaded
+            + expires: `2019-10-30T11:44:33+00:00` (string) - Expiration time of the temporary upload url
```
mrtus: Q: Shouldn't this be returning a file id instead?
```suggestion
            + type: `file` (string)
            + id: `<uuid>` (string) 
```

JordyDeGroeve (dismissed):

JordyDeGroeve (commented):
Did not mean to approve. Let's address the remarks first

apiary.apib
Jdww: Nope, the way the file upload works is it returns the temporary url where the proposed file can be uploaded

mrtus: Ah right, forgot about that 👍

mrtus (approved):

MathiasDL (approved):

JordyDeGroeve (approved):

rainbow-alex (changes requested):
The subject type in .info has more options than in .list. Should be the same everywhere.

Jdww:
> The subject type in .info has more options than in .list. Should be the same everywhere.

The .list has a required filter with a limited amount of subject types. I understand that it would be hard to get the .info of something not returned in .list but it's possible. People can use api v1 and migrate the id. Not sure what would be the correct approach. We could define more subject types in .list but they will not be returned until we add those filters.

rainbow-alex (approved):

tiagobastos:
Didn't we forget about an entry on the change log?

PR #639, opened 2022-06-22
Add grouped lines to subscriptions.info

JordyDeGroeve:


src/05-invoicing/subscriptions.apib
```diff
@@ -139,6 +139,37 @@
                         + rate: 0.21 (number)
                         + taxable (Money)
                         + tax (Money)
+            + grouped_lines (array)
```
Jdww, line 142: PP: Could we add a data structure for this? Or is it different for invoices

src/05-invoicing/subscriptions.apib
```diff
@@ -139,6 +139,37 @@
...
+                + (object)
+                    + section (object)
+                        + title (string)
```
Jdww, line 145: Title can be null on invoices.info so I assume it will be the same here

src/05-invoicing/subscriptions.apib
JordyDeGroeve: There is some difference indeed 😞  (withheld tax f.e.).
On top of that, there is quite a difference between the request and response structure (apib-wise) so it might become confusing. Gonna keep it like it is, to avoid messing up the documentation on several endpoints at once.

src/05-invoicing/subscriptions.apib
JordyDeGroeve: It is [not nullable](https://github.com/teamleadercrm/api/blob/master/src/05-invoicing/invoices.apib#L158) on invoices.info 🤔

Jdww: The documentation is wrong, it can return null.

joaoinacio:
Closing this one for now, we'll re-create it with the fully updated docs.

PR #649, opened 2022-08-29
ATL-1665 Subscriptions.info - book_invoices and autosend

Jdww:
We want to return book_invoices and autosend in subscriptions.info

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -399,6 +399,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 - We added `invoices.removePayment` and `invoices.removePayments`.
 - We added `department_id` to the filters on `subscriptions.list`.
+- We added `book_invoices` and `autosend` to `subscriptions.info`.
 
 ### July 2022
 
@@ -3504,6 +3505,10 @@ Get details for a single subscription
                         + taxable (Money)
                         + tax (Money)
             + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
+            + `book_invoices`: true (boolean)
+            + autosend (object)
+                + enabled: true (boolean)
+                + via_postal_service: true (boolean)
 
 ### subscriptions.deactivate [POST /subscriptions.deactivate]
 
--- a/src/05-invoicing/subscriptions.apib
+++ b/src/05-invoicing/subscriptions.apib
@@ -145,6 +145,10 @@ Get details for a single subscription
                         + taxable (Money)
                         + tax (Money)
             + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
+            + `book_invoices`: true (boolean)
+            + autosend (object)
+                + enabled: true (boolean)
+                + via_postal_service: true (boolean)
 
 ### subscriptions.deactivate [POST /subscriptions.deactivate]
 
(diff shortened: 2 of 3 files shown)


joaoinacio (approved):

lowiebenoot (approved):

apiary.apib
```diff
@@ -3504,6 +3505,10 @@ Get details for a single subscription
                         + taxable (Money)
                         + tax (Money)
             + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
+            + `book_invoices`: true (boolean)
```
mrtus, line 3508: In hindsight, I think I would've loved to see this as `invoice_status: draft|booked` instead, booleans are evil :).

I think we should use enums because "true" or "false" does not say much about what the other state exactly is. What is the state of the invoice if `book_invoices` is not set to `true`, by using enums we would document our API through "code".

Jdww: You don't know the status of the invoices created from a subscription. You only know the initial state after creation. I think this suggestion would make it more confusing.

mattkomarnicki (approved):

apiary.apib
mrtus: That's what I am talking about @Jdww, we are talking about the same thing.

We want to decide what the initial state of an invoice must be, but what is the state of that invoice if `book_invoices` is set to `false`? We know that it is booked when set to true, but what is the false state? You could say "draft", but how do you know, it's not reflected as such.

joaoinacio: > That's what I am talking about @Jdww, we are talking about the same thing.
> 
> We want to decide what the initial state of an invoice must be, but what is the state of that invoice if `book_invoices` is set to `false`? We know that it is booked when set to true, but what is the false state? You could say "draft", but how do you know, it's not reflected as such.

We could also say that booking is an action, and it doesn't really matter which other states there are (and we don't support selecting any other state(s)).
So, this is **not** about selecting an initial invoice state, it's about whether the invoices should be booked after creation (or not).

Our own statement on actions even mentions `invoices.book` as an example:

> Our API consists of HTTP RPC-style methods, in the form of https://api.focus.teamleader.eu/resource.action.We chose this action based approach over the more popular REST, because it enables us to have domain related actions on resources such as invoices.book ...

joaoinacio:
✅  Manually checked.

src/05-invoicing/subscriptions.apib
```diff
@@ -145,6 +145,10 @@ Get details for a single subscription
                         + taxable (Money)
                         + tax (Money)
             + `web_url`: `https://focus.teamleader.eu/subscription_detail.php?id=e2314517-3cab-4aa9-8471-450e73449041` (string)
+            + `book_invoices`: true (boolean)
+            + autosend (object)
+                + enabled: true (boolean)
+                + via_postal_service: true (boolean)
```
mrtus, line 151: As discussed, the structure above will be converted to the structure below.

The email sending method will always be explicitly required as this is currently a behaviour that cannot be disabled. The sending methods list will only be required when `book_and_send` is configured as action.

```json
{
    "invoice_generation": {
        "action": "draft|book|book_and_send",
        "sending_methods": [
            {
                "method": "email"
            },
            {
                "method": "postal_service"
            }
        ]
    }
}

```

apiary.apib
mrtus: Conclusion: https://github.com/teamleadercrm/api/pull/649#discussion_r958645094

PR #655, opened 2022-09-14
Add documentation for (updating) collections

joaoinacio:


mattkomarnicki (approved):

apiary.apib
```diff
@@ -166,6 +166,47 @@ Common used status codes:
  - `429` - Too Many Requests, your client has reached the API rate limit
  - `500` - Internal Server Error, something went wrong on our end
 
+### Collections
+
+Collections are groups of items, represented as JSON arrays `[]`, such as:
```
mrtus, line 171: Q: Maybe we can make this more open?

```suggestion
Collections are groups of items, represented as JSON arrays `[]`, such as, but not limited to:
```

Jdww: Isn't that already implied with such as? Feels like you are mixing it up with an expression like including but not limited to.

mrtus: If that is the case @Jdww sure 👍

tiagobastos (approved):

Jdww (approved):

PR #658, opened 2022-09-14
[ATL-591] As an API User, I can send an invoice

mattkomarnicki:
### Link to issue 

https://teamleader.atlassian.net/browse/ATL-591


apiary.apib
```diff
@@ -3141,6 +3142,52 @@ Update a draft invoice. Booked invoices cannot be updated.
...
+                       + type (enum, required)
+                           + Members
+                               + company
+                               + contact
+                       + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+                   + email: `info@teamleader.eu` (string, required)
+       + attachments: `e76d381a-f7ea-04d1-9a32-14aab27bc1e2` (array[string], optional) - an array of file ids
+       + allow_reminders: true (boolean, optional) - If set to false, the endpoint will return a 400 response if invoice was already sent
+           + Default: true
+       + attach_invoice_pdf: true (boolean, optional) - If set to true, the invoice's pdf will be included as an attachment to the email
```
mrtus, line 3187: Q: We do not require such property on `/quotations.send`, why do we require this here? Seems like we have a behavioural difference we'll need to adjust first.

This is also a `/invoices.send` endpoint, I'm not sure I understand why the PDF would have to be optional?

apiary.apib
```diff
@@ -3141,6 +3142,52 @@ Update a draft invoice. Booked invoices cannot be updated.
...
+                + (object)
+                   + customer (object, optional, nullable)
+                       + type (enum, required)
+                           + Members
+                               + company
+                               + contact
+                       + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+                   + email: `info@teamleader.eu` (string, required)
+       + attachments: `e76d381a-f7ea-04d1-9a32-14aab27bc1e2` (array[string], optional) - an array of file ids
+       + allow_reminders: true (boolean, optional) - If set to false, the endpoint will return a 400 response if invoice was already sent
```
mrtus, line 3185: Q: Why a 400 response? Wouldn't a 403 or 409 be a more fitting status code?

> the endpoint will return a 400 response if invoice was already sent

A 400 status hints to the user that they can adjust their request to resolve the problem, this is not the case if I'm reading the description of this property correctly.

Which means 400 is not correct, so probably 403 is the most fitting status code here.

mrtus, line 3185: From past discussions we will not be working with booleans anymore in our API request & responses.

Let's find a better name & the values to reflect this behaviour.

e.g. (albeit let's think about the not-sending-part value)
```json
{
  "(send_)reminders?": "allow | (forbid?|disallow)"
}
```

What exactly does this property facilitate?

apiary.apib
```diff
@@ -3141,6 +3142,52 @@ Update a draft invoice. Booked invoices cannot be updated.
...
+
+Send an invoice via e-mail.
+
++ Request (application/json)
+    + Attributes (object)
+        + id: `d885e5d5-bacb-4607-bde9-abc4a04a901b` (string, required)
+        + content (object, required)
+             + subject: `Invoice` (string, required)
+             + body: `Please find your invoice attached to this mail` (string, required)
+             + template_id: `045cd6a9-7527-09c9-aa16-fd5ac7953e71` (string, optional, nullable)
```
mrtus, line 3155: Q: What is the difference between quotations & invoices?
Aren't quotations also send with templates?

Can you explain why is there a template id option here and not on quotations?

apiary.apib
mrtus, line 3187: Same as the other property, let's at least update this to a non-boolean value.

e.g. (albeit let's maybe think about the not-sending-part value)
```json
{
  "(invoice_)pdf_file": "attach | do_not_send"
}
```

apiary.apib
mrtus: This structure is also not matching the structure we documented on`/quotations.send`, there it isn't wrapped in `content` at all.

Let's make this consistent by updating the structure here or on the just published `/quotations.send`

apiary.apib
joaoinacio: Not sure I understand, it's not required but optional.
AFAIK the pdf may not be attached if f.e. it will be instead made available in invoicecloud. 

> Seems like we have a behavioural difference we'll need to adjust first.

Yes, but invoices and quotations are not the same so having slightly different behaviour is to be expected, no?

mrtus: > Not sure I understand, it's not required but optional.

I meant having the property available.

mattkomarnicki:
After internal discussion, all PRs linked to [the related ticket](https://teamleader.atlassian.net/browse/ATL-591) will be closed. These are:

1. https://github.com/teamleadercrm/core/pull/29059
2. https://github.com/teamleadercrm/api-internal/pull/430
3. https://github.com/teamleadercrm/api/pull/658 (feedback provided here requires us to move the ticket to the left and go back to the whiteboard).

Closing this now. cc @joaoinacio

PR #676, opened 2023-02-20
[TIDAL-3804] Undo changes to `invoices.list` filters

adognameddog:
These were incorrectly changed to use a `type` and `id`.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2869,28 +2869,22 @@ Get a list of invoices.
 
     + Attributes (object)
         + filter (object, optional)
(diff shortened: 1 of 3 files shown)


mrtus (commented):
👍

adognameddog:
@mrtus In the request, in the `filter` we use `entity_id`: `some uuid`. In the _response_, we return those as type/id.

mrtus (approved):

driesd (approved):

lorgan3 (approved):

lorgan3:
✅ checked

PR #684, opened 2023-03-23
Add missing migrate types

JordyDeGroeve:


tiagobastos (dismissed):
Should we add this change to the backlog of changes?

joaoinacio (dismissed):

apiary.apib
```diff
@@ -5207,6 +5208,9 @@ Translates an ID from the deprecated API into a new UUID.
                 + creditNote
                 + subscription
                 + quotation
+                + timeTracking
+                + deliveryNote
```
mrtus: Don't have any api yet for this, so no need to document it yet.

JordyDeGroeve: Indeed. I'll remove it.

mrtus (dismissed):

apiary.apib
```diff
@@ -5262,6 +5267,8 @@ Translates an ID from the deprecated API into a new UUID.
                     + creditNote
                     + subscription
                     + quotation
+                    + timeTracking
+                    + customField
```
mrtus, line 5271: Q: Should be `customFieldDefinition` I think?

JordyDeGroeve: It is.. 🙈

apiary.apib
```diff
@@ -424,6 +424,7 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### March 2023
 
+- We added `customField`,`timeTracking` and `department` as allowed types on `migrate.id`.
```
tiagobastos, line 427: Now, the change log doesn't match the added types.

mrtus (approved):

apiary.apib
JordyDeGroeve: It does. There was not really a concept like CF definition/value in API v1. Hence it gets transposed from `customField` to `customFieldDefinition` in the migrate call.

tiagobastos (approved):

JorenSaeyTL (approved):

PR #691, opened 2023-05-09
Filter team lead

mrtus:
We want to find which teams a user has access to as a team lead

### Added
- Allow `/teams.list` to be filtered on by `team_lead_id`

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -422,9 +422,12 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+### May 2023
+- We added `team_lead_id` to the filters on `teams.list`.
+
 #### April 2023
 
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -422,9 +422,12 @@ We consider the following changes to be backwards-incompatible:
 We list all backwards-compatible additions here. These are currently available in all published versions.
 (There is also a [list of backwards-incompatible upgrades](#changelog) available, but those only apply if you [upgrade your API version](#upgrading-your-api-version).)
 
+### May 2023
+- We added `team_lead_id` to the filters on `teams.list`.
+
 #### April 2023
```
StephenBeirlaen, line 428: Shouldn't this be

```suggestion
### April 2023
```

?

StephenBeirlaen (approved):

apiary.apib
mrtus: Seems like the opposite actually, all should be heading 4 🙈

mrtus: https://github.com/teamleadercrm/api/pull/692

aravind-teamleader (approved):

KristofColpaert (approved):

PR #714, opened 2023-07-17
Add `currencies.exchangeRates` endpoint

pidera:


apiary.apib
```diff
@@ -1426,6 +1427,27 @@ Get a list of all document templates.
...
+
++ Response 200 (application/json)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + code: `USD` (Currency)
+                + symbol: `$` (string)
+                + name: `US Dollar` (string)
+                + exchange_rate: `1.1238` (number)
```
mrtus: Maybe one information point I am missing here, I think, is this exchange rate 1.1238 dollar to 1 euro, or is it the other way around?

Not sure how we can make this more clear in the API contract, or if we can simply get away with adding this as a comment on the documentation.

mrtus (approved):

Jdww (approved):

Jdww:
Manual checked ✅

uhajzeraj (approved):

PR #715, opened 2023-07-19
Add `invoice_currency` on `invoices.draft`and `invoices.update`

pidera:


apiary.apib
```diff
@@ -3400,6 +3404,9 @@ Update a draft invoice. Booked invoices cannot be updated.
         + expected_payment_method (ExpectedPaymentMethod, optional, nullable)
         + custom_fields (array[CustomFieldValue], optional)
         + document_template_id: `179e1564-493b-4305-8c54-a34fc80920fc` (string, optional)
+        + invoice_currency (object, optional)
```
mrtus: Q: Isn't the `invoice` prefix redundant since we are already on an invoice resource?

I think solely `currency` or `price_currency` might be better?

@Miljar what are your thoughts?

```suggestion
        + currency (object, optional)
```

mrtus: Maybe we should expand on what the default would be if not given?

```suggestion
        + invoice_currency (object, optional) - Defaults to department currency when not given
```

Miljar: I definitely need more context give an opinion. Is there a JIRA ticket to read?

pidera: Yes, sure: https://teamleader.atlassian.net/browse/ATL-2189 and https://teamleader.atlassian.net/browse/ATL-2193

Miljar: Firstly, I'd decouple "currency" & "exchange rate" and not put in in an overarching object "invoice currency" (reasoning: a currency has a couple of properties, but "exchange rate" is not one of them since that's between 2 currencies)

Secondly, I don't understand why you need to provide an exchange rate. It's not clear which currencies you're converting between. I probably don't know enough about the invoicing domain, but why can someone define their own exchange rate, if we have that information somewhere in the system? (It's stricken through in the JIRA ticket, but I don't know why)

Jdww: exchange rate changes daily? Why would someone store it in teamleader

Jdww (dismissed):

Jdww (approved):

pidera:
Closing, will need BC layer and version release update first, see [ATL-2189](https://teamleader.atlassian.net/browse/ATL-2189)

[ATL-2189]: https://teamleader.atlassian.net/browse/ATL-2189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #717, opened 2023-07-24
Property `currency` on `item_lines` for `invoices.draft` and `invoices.update` is now optional

pidera:


tiagobastos (approved):

apiary.apib
```diff
@@ -3335,7 +3338,8 @@ Draft a new invoice.
                         + extended_description: `Some more information about this awesome product` (string, optional, nullable) - Uses Markdown formatting
                         + unit_of_measure_id: `f79d3e04-b8dc-0637-8f18-ca7c8fc63b71` (string, optional, nullable)
                         + unit_price (object, required)
-                            + Include Money
+                            + amount: `123.30` (number, required)
+                            + currency: `EUR` (Currency, optional)
```
mrtus, line 3342: We should probably make clear which currency it will then rely on I think?

I'm not entirely sure if we really need to make this optional too. Mostly because a price always exists out of an amount & currency.

What was the reasoning to make this optional?

pidera: Reasoning for this change can be found on this ticket: https://teamleader.atlassian.net/browse/ATL-2203

Jdww (approved):

apiary.apib
mrtus: Not entirely convinced about that reasoning but can't find any counter arguments either.

mrtus (approved):

pidera:
Won't do for now, superseded by [ATL-2244](https://teamleader.atlassian.net/browse/ATL-2244)

[ATL-2244]: https://teamleader.atlassian.net/browse/ATL-2244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #718, opened 2023-07-25
Invoices.credit - Partial credit note

Jdww:


mrtus (approved):

src/05-invoicing/invoices.apib
```diff
@@ -446,6 +446,14 @@ Credit an invoice completely.
     + Attributes (object)
         + id: `d885e5d5-bacb-4607-bde9-abc4a04a901b` (string, required)
         + credit_note_date: `2016-02-04` (string, optional)
+        + partial (object, optional)
```
mrtus, line 449: Q: Not sure if we really need the root attribute though. Wouldn't it suffice to document that if not given the invoice is creditted in full? If lines are given, lines are taken?

Jdww: The root attribute was added because the grouped lines and discounts belong together.

PR #723, opened 2023-08-14
Release quotation expiry

JordyDeGroeve:


src/03-deals/quotations.apib
```diff
@@ -43,8 +43,12 @@ Get a list of quotations.
                     + Members
                         + open
                         + accepted
+                        + expired
                         + rejected
                         + closed
+                + expiry (object) - returned if user has access to quotation expiry and when `includes=expiry` is requested
+                        + expires_after: `2023-04-05` (string, nullable)
+                        + lock_after_expiry (boolean)
```
joaoinacio, line 51: 🤨

pidera (dismissed):

pidera (approved):

apiary.apib
```diff
@@ -2810,6 +2827,21 @@ Delete a quotation.
...
+Request an extension for a quotation. Only possible for expired quotations.
+
++ Request (application/json)
+
+    + Attributes (object)
+        + id: `d885e5d5-bacb-4607-bde9-abc4a04a901b` (string, required)
+        + send_to: `info@teamleader.eu` (string, required)
+        + send_copy_to: maria@teamleader.eu` (string, optional, nullable)
+        + subject: `I missed the deadline' quotation` (string, required)
+        + content: `Can you reopen this quotation, please?` (string, required)
```
Jdww, line 2841: PP: The example text is confusing, this looks like a request an end customer would do to our customer. However end customers will not call the api.

JordyDeGroeve: But this is exactly what this endpoint does. Our customer will never request an extension, it will update the expiration date in that case. This is for extensions request by the end customer.

Maybe this endpoint should be reconsidered (to document publicly)

PR #739, opened 2023-09-26
Meetings 

Jdww:


tiagobastos (dismissed):

joaoinacio (dismissed):

joaoinacio (dismissed):

tiagobastos (dismissed):

tiagobastos (dismissed):

tiagobastos (dismissed):

joaoinacio (dismissed):

tiagobastos:
✔️ Manually checked

gdegirmenci (dismissed):

apiary.apib
```diff
@@ -3315,6 +3338,143 @@ Get information about a meeting.
...
+                        + area_level_two_id: `db232cf8-ad4a-024b-941f-15a7a74f0fd2`
+                + Properties
+                    + type: `customLocation` (string, required)
+                    + id: `4191c526-5e26-0818-8e16-9523215b5081` (string, required)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + custom_fields (array[CustomFieldValue], optional)
+
++ Response 201
+
```
mrtus: Q: Response returns an id of the event does it not?

We should probably document that too.

mrtus (dismissed):

src/04-calendar/meetings.apib
```diff
@@ -165,4 +187,146 @@ Get information about a meeting.
...
+                    + id: `4191c526-5e26-0818-8e16-9523215b5081` (string, required)
+        + milestone_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + deal_id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string, optional)
+        + custom_fields (array[CustomFieldValue], optional)
+
++ Response 201 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + type: `meeting` (string)
```
mrtus, line 257: It creates an `event` 😉 
```suggestion
            + type: `event` (string)
```

tiagobastos: The endpoint is `/meetings.schedule` it returns a meeting UUID.

mrtus: Ahjajajaj its the meeting creation endpoint, nevermind! I was think about `/tasks.schedule` 🤦

mrtus (dismissed):

joaoinacio (dismissed):

tiagobastos (dismissed):

joaoinacio (dismissed):

gdegirmenci (dismissed):

apiary.apib
```diff
@@ -3228,7 +3233,9 @@ Get a list of meetings.
                     + Members
                         + open
                         + done
-
+                + recurrence (object, nullable)
+                    + type: `recurrence` (string)
```
mrtus, line 3243: Q: We don't expose this resource anywhere yet, do we?

To which type refers this id then? Without exposing the resource I think we shouldn't include this property yet.

mrtus (dismissed):

apiary.apib
joaoinacio: > Q: We don't expose this resource anywhere yet, do we?
> 
> To which type refers this id then? Without exposing the resource I think we shouldn't include this property yet.

🤔 Right. the goal is to be able to list recurring meetings from a recurrence id, so providing the recurrence (definition) isn't necessary for now.
We can omit the type property for now and provide just the id, which is the necessary info to have.

gdegirmenci (dismissed):

apiary.apib
mrtus: Okay, where would that id then come from? From existing meetings which return the recurrence?

Though, I wouldn't really remove the type if we'd still expose the id.

joaoinacio (approved):

mrtus (approved):

tiagobastos (approved):

PR #740, opened 2023-10-03
Calls endpoints

Jdww:


apiary.apib
```diff
@@ -429,7 +429,8 @@ We list all backwards-compatible additions here. These are currently available i
 ### September 2023
 - We added the `meetings.list` and `meetings.info` endpoints.
 - We added `time_estimated` and `amount_unbilled` to `projects-v2/projects.list`.
-- We added sorting on `amount_billed`, `amount_paid`, `amount_unbilled`, `external_budget_spent`, `external_budget`, `internal_budget` and `time_estimated` to `projects-v2/projects.list`. 
+- We added sorting on `amount_billed`, `amount_paid`, `amount_unbilled`, `external_budget_spent`, `external_budget`, `internal_budget` and `time_estimated` to `projects-v2/projects.list`.
+- We added the `calls.add` endpoint.
```
joaoinacio: This should be in October

joaoinacio (dismissed):

tiagobastos (dismissed):

tiagobastos:
✔️ Manually checked

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+            + responsible_user (object, nullable)
+                + type: `user` (string)
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
+            + scheduled_at: `2016-02-04T16:00:00+00:00` (string)
+            + status: `done` (enum[string])
+                + Members
+                    + `open`
+                    + `done`
+
+### calls.complete [GET /calls.complete]
```
JordyDeGroeve:
```suggestion
### calls.complete [POST /calls.complete]
```

JordyDeGroeve (dismissed):

BrechtBonte (dismissed):

gdegirmenci (dismissed):

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+    + Attributes
+        + description: `Description of the call` (string, optional)
+        + customer (object, required)
+            + type (enum, required)
+                + Members
+                    + company
+                    + contact
+            + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+        + due_date: `2016-02-04T16:00:00+00:00` (string, required)
+        + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22` (string, required)
```
mrtus: Q: I'm wondering if we shouldn't allow to be open for change in the future where we would allow a team to be assigned, just like tasks?


```suggestion
        + responsible: `98b2863e-7b01-4232-82f5-ede1f0b9db22` (object, required)
            + type (enum, required)
                + Members
                    + user
            + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
```

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+                + optional_contact (object, nullable)
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                    + type: `contact` (string)
+                + outcome (object, nullable)
+                    + type: `callOutcome` (string)
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + outcome_summary: `Called, but was not available` (string, nullable)
+                + responsible_user (object, nullable)
+                    + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
+                    + type: `user` (string)
```
mrtus: Q: Let's open it up for eventually teams to be assigned?

```suggestion
                + responsible (object, nullable)
                    + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
                    + type: `user` (string)
```

apiary.apib
mrtus: Q: Should we talk about assignee or responsible?

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+        + id: `6fac0bf0-e803-424e-af67-76863a3d7d16` (string, required)
+
++ Response 200 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+            + added_at: `2016-02-01T16:44:33+00:00` (string, nullable)
+            + completed_at: `2016-02-04T16:44:33+00:00` (string, nullable)
+            + customer (object, nullable)
```
mrtus: Q: Not everyone you call is already a customer. What about `recipient`?

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+                + type: `company` (string)
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
+            + description: `Description of the call` (string, nullable)
+            + optional_contact (object, nullable)
+               + type: `contact` (string)
+               + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
```
mrtus: Q: Can I assume that the optional contact is part of the "customer" you are calling?

Reasoning, I call a company where I have a contact, which I assume `optional_contact` is for?
From what I see this contact when added to the call is automatically linked to the company the call is for too, so this would make sense to have it structured like 👇 

They should probably be grouped:
```suggestion
            + recipient (object, nullable)
               + customer: (object, required)
                  + type: `company` (string)
                  + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
               + for_attention_of: (object, nullable)
                  + type: `contact` (string)
                  + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
            + description: `Description of the call` (string, nullable)
```

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+                + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+            + outcome_summary: `Called, but was not available` (string, nullable)
+            + responsible_user (object, nullable)
+                + type: `user` (string)
+                + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
+            + scheduled_at: `2016-02-04T16:00:00+00:00` (string)
+            + status: `done` (enum[string])
+                + Members
+                    + `open`
+                    + `done`
```
mrtus: Q: If the endpoint is going to be called `calls.complete` the status should probably be `completed` instead of `done`? Especially since it is also `completed_at`.

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + start_date: `2021-11-21` (string) - Filter on dates that are on or after the given day
+            + end_date: `2022-05-13` (string)  - Filter on dates that are on or before the given day
+            + relates_to (object, optional) - Filter calls by related object (currently only companies allowed)
+                + id: `415fa7e0-e319-0952-b973-f067f2b49d56` (string)
+                + type: `company` (string)
+            + milestone_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Filter calls by milestone
```
mrtus: Q: Let's drop milestone since this is no longer a thing in next-gen?
Unless it is a requirement then 👍

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                    + type: `contact` (string)
+                + outcome (object, nullable)
+                    + type: `callOutcome` (string)
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + outcome_summary: `Called, but was not available` (string, nullable)
+                + responsible_user (object, nullable)
+                    + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
+                    + type: `user` (string)
+                + scheduled_at: `2021-11-21T19:14:15+00:00` (string)
```
mrtus, line 3367: Q: Is this the time for when the call should take place? If that is the case I think it must be `on` or `for` and never `at`.

apiary.apib
```diff
@@ -3315,6 +3317,151 @@ Get information about a meeting.
...
+
+    + Attributes
+        + description: `Description of the call` (string, optional)
+        + customer (object, required)
+            + type (enum, required)
+                + Members
+                    + company
+                    + contact
+            + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string, required)
+        + due_date: `2016-02-04T16:00:00+00:00` (string, required)
```
mrtus: We only talk about `date` when talking about a day without referencing the time, since we are also talking about a time, this thus must be `due_on`

```suggestion
        + due_on: `2016-02-04T16:00:00+00:00` (string, required)
```

On the other hand, we should pull this in one line with the response property `scheduled_on`. Given that I think we can go with `due_on` instead of `scheduled_on` or `_for` I'm not sure which one is the most correct term to be used in this case.

apiary.apib
tiagobastos: Already removed in the follow-up.
https://github.com/teamleadercrm/api/pull/749/commits/d8d1ede67cf97d5ca8bf727d581ab165daf0b620

apiary.apib
tiagobastos: We can't add non-TL customers to the calls so they are customers.

mrtus: I understand but a contact / company is not automatically a customer, it could be a lead or a partner or ...

JordyDeGroeve (dismissed):

apiary.apib
pidera: As an addition, if we look at a similar structure on `invoices.list`, should we group the 'type' and 'id' of `for_attention_of` in a `contact` object? This would open up the options to add expose additional properties in the future.

```suggestion
            +recipient (object, nullable)
               + customer: (object, required)
                  + type: `company` (string)
                  + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
               + for_attention_of: (object, nullable)
               	  + contact (object, required)
                    + type: `contact` (string)
                    + id: `2659dc4d-444b-4ced-b51c-b87591f604d7` (string)
            + description: `Description of the call` (string, nullable)
```

adognameddog (dismissed):

joaoinacio (dismissed):

JordyDeGroeve (dismissed):

joaoinacio (dismissed):

apiary.apib
mrtus: @tiagobastos this was marked as resolved, but no conclusion was provided?

I still think this property is not correctly named, especially since we learned that calls can be about incoming ones, which makes it no longer scheduled, but an occurrence in the past.

tiagobastos (dismissed):

src/04-calendar/calls.apib
```diff
@@ -0,0 +1,158 @@
...
+
+### calls.list [POST /calls.list]
+
+Get a list of calls.
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + start_date: `2021-11-21` (string) - Filter on dates that are on or after the given day
+            + end_date: `2022-05-13` (string)  - Filter on dates that are on or before the given day
```
mrtus: Q: On which time property do these filter? Added / scheduled / completed at? We must narrow the naming down to the behaviour, as an API user this is unclear currently.

apiary.apib
joaoinacio: > Q: Is this the time for when the call should take place? If that is the case I think it must be on or for and never at.

We use `on` for dates and `at` for date+time, so I don't see any issue in the naming here.

> I still think this property is not correctly named, especially since we learned that calls can be about incoming ones, which makes it no longer scheduled, but an occurrence in the past.

Calls are still mostly scheduled in the future, don't see how the name is incorrect?

src/04-calendar/calls.apib
tiagobastos: The description could be made clearer. 
How about: `Filter on calls occurring on or before/after a given date`

mrtus: I think these time based filters must then be named as `scheduled_after` and `scheduled_before` (assuming we'd keep the `scheduled_at` naming), just like we have this format on other filters with `created_after|before`, on invoices we have`invoice_date_after|before` & ..., for time tracking we have `started_after|before`, `ended_after|before` & ...

The naming convention to these filters is that they must be as close as possible to the info property, so if something is named `scheduled_at` it must be filterable as `scheduled_x`, if that makes sense?

tiagobastos: https://github.com/teamleadercrm/core/pull/35891

tiagobastos (approved):

joaoinacio (approved):

mrtus (approved):

apiary.apib
mrtus: Topic was discused on internal channels.

I'm okay if we stay on `scheduled_at`

pidera (approved):

PR #743, opened 2023-10-13
ATL-117 Calls and Call outcome

Jdww:
Previous: https://github.com/teamleadercrm/api/pull/740

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -431,7 +431,7 @@ We list all backwards-compatible additions here. These are currently available i
 - We added `time_estimated` and `amount_unbilled` to `projects-v2/projects.list`.
 - We added sorting on `amount_billed`, `amount_paid`, `amount_unbilled`, `external_budget_spent`, `external_budget`, `internal_budget` and `time_estimated` to `projects-v2/projects.list`.
 - We added the `calls.add` endpoint.
-- We added the `calls.info` endpoint.
+- We added the `calls.list` and `calls.info` endpoint.
 
 ### August 2023
 - We added `quotation` as a type for `cloudPlatforms.url`.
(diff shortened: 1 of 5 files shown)


apiary.apib
```diff
@@ -3303,6 +3307,106 @@ Add a new call.
...
 
+### calls.list [GET /calls.list]
+
+Get a list of calls.
+
++ Request (application/json)
+    + Attributes (object)
+        + filter (object, optional)
+            + milestone_id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, optional) - Filter calls by milestone
+            + employee_id: `b570e8c4-6c62-0f7a-a55b-6f330a127410` (string, optional) - Filter calls by user
```
mrtus: This must be `user_id`, or even better would be `responsible_user_id` since that is the term we are using in the responses
```suggestion
            + responsible_user_id: `b570e8c4-6c62-0f7a-a55b-6f330a127410` (string, optional) - Filter calls by user
```

Jdww: Removed

mrtus (approved):

joaoinacio (approved):

apiary.apib
```diff
@@ -3303,6 +3307,105 @@ Add a new call.
...
+                + Default: `20`
+            + number: `1` (number, optional) - The current page, the first page is 1
+                + Default: `1`
+
++ Response 200 (application/json)
+
+    + Attributes (object)
+        + data (array)
+            + (object)
+                + id: `13` (number)
```
tiagobastos: B: We don't have integer identifiers. Same for info.

apiary.apib
```diff
@@ -3303,6 +3307,105 @@ Add a new call.
...
+                + description: `Call with customer` (string, nullable)
+                + scheduled_at: `2021-11-21T19:14:15+00:00` (string)
+                + responsible_user (object, nullable)
+                    + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string)
+                    + type: `user` (string)
+                 + outcome (object, nullable)
+                    + type: `callOutcome` (string)
+                    + id: `32665afd-1818-0ed3-9e18-a603a3a21b95` (string)
+                + outcome_summary: `Called, but was not available` (string, nullable)
+                + status: `open` (string)
```
tiagobastos: Shouldn't this be an enum of available status? 
Same for info.

tiagobastos (changes requested):
nitpick: Sort the `.list` & `.info` alphabetically so it's easier to identify which fields are missing/added when comparing both.

tiagobastos (approved):

tiagobastos:
✔️ Manually checked

PR #747, opened 2023-10-30
Document | `/calls.complete`

tiagobastos:
Previous: ~https://github.com/teamleadercrm/api/pull/743~ https://github.com/teamleadercrm/api/pull/740

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -428,7 +428,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### October 2023
 - We added the `calls.list` and `calls.info` endpoint.
-- We added the `callOutcome.list` endpoint.
+- We added the `callOutcomes.list` endpoint.
+- We added the `calls.complete` endpoint.
 
 ### September 2023
 - We added the `meetings.list` and `meetings.info` endpoints.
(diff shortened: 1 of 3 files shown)


apiary.apib
```diff
@@ -428,7 +428,8 @@ We list all backwards-compatible additions here. These are currently available i
 
 ### October 2023
 - We added the `calls.list` and `calls.info` endpoint.
-- We added the `callOutcome.list` endpoint.
+- We added the `callOutcomes.list` endpoint.
+- We added the `calls.complete` endpoint.
```
mrtus, line 432: Q: Do we "complete" a call, or do we "end" or maybe "finish" a call?

Not sure the term "complete" is the best fit here.

mrtus (approved):
👌 although I'm doubting about the action name.

apiary.apib
joaoinacio: > Q: Do we "complete" a call, or do we "end" or maybe "finish" a call?
> 
> Not sure the term "complete" is the best fit here.

🤔 aren't those synonyms though?

I think when it was introduced one of the goals was also to have consistency between all the events endpoints (calls.complete, tasks.complete, meetings.complete) even if the term may not be the perfect for each 🤷‍♂️

joaoinacio (approved):

PR #752, opened 2023-11-13
Change type due to returning code: 405000

basven1997:
The method GET is not supported in your api. Only the post is supported in the API


mrtus (commented):
Good catch!

The rendered file [`apiary.apib`](https://github.com/teamleadercrm/api/blob/master/apiary.apib) also needs to be updated.
You can do this through `make build` or if that does not work, you can do the change manually in this case.

Once that is done we can get this merged.

basven1997:
Maybe its better you teamleader will fix this item.

mrtus:
Sure @basven1997, you can find the changes here https://github.com/teamleadercrm/api/pull/753

PR #756, opened 2023-11-15
Add missing `new` status on `deals.list` and `deals.info`

lowiebenoot:
The `new` status was missing on the `status` attribute on the response of `deals.info` and on the `status` filter on `deals.list`.

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -2249,7 +2249,7 @@ Get a list of deals.
                 + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22`, `98b2863e-7b01-4232-82f5-ede1f0b9db24` (array[string], optional)
             + updated_since: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
             + created_before: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
-            + status: `open`, `won`, `lost` (array[string], optional)
+            + status: `new`, `open`, `won`, `lost` (array[string], optional)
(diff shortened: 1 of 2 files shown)


apiary.apib
```diff
@@ -2249,7 +2249,7 @@ Get a list of deals.
                 + responsible_user_id: `98b2863e-7b01-4232-82f5-ede1f0b9db22`, `98b2863e-7b01-4232-82f5-ede1f0b9db24` (array[string], optional)
             + updated_since: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
             + created_before: `2018-02-11T16:45:30+00:00` (string, optional) - This date is inclusive
-            + status: `open`, `won`, `lost` (array[string], optional)
+            + status: `new`, `open`, `won`, `lost` (array[string], optional)
```
mrtus: Code wise the new filter is not allowed, so that will require some changes too

lowiebenoot: Ah, I didn't check the validator. Indeed not allowed at the moment, which doesn't really make sense I guess?

lowiebenoot: Actually, filtering on `open` also returns `new` deals...

mrtus: Oh, then I'm not sure if we need to include the new status on the filter? 🤔

lowiebenoot: Yeah, i'll drop it on the filter

mrtus (approved):

lowiebenoot:
Closing, as we will further discuss whether we should have `new` as a status or not.

PR #761, opened 2023-11-24
notes.add

Jdww:


src/01-general/notes.apib
```diff
@@ -23,4 +23,28 @@ Get a list of all notes.
...
+                    + Members
+                        + `user`
+                + id: `f29abf48-337d-44b4-aad4-585f5277a456` (string, required)
+
++ Response 201 (application/json)
+
+    + Attributes (object)
+        + data (object)
+            + type: `note` (string)
+            + id: `ee94b4c0-5786-0517-9d26-8dd1e6406f20` (string)
```
gdegirmenci: eol

apiary.apib
```diff
@@ -1567,6 +1567,30 @@ Get a list of all notes.
                     + id: `36386b05-936e-4cc0-9523-bd20d797ebf5` (string)
                 + added_at: `2016-01-01T00:00:00+00:00` (string)
 
+### notes.add [POST /notes.add]
```
mrtus: Following the reasoning from `create` vs `add`, this will have to be `create`

Create => doesn't exist out of TLF
Add => does exist out of TLF

We could argue a note can exist out of TLF but I think create is the more applicable here.

joaoinacio (approved):

joaoinacio (approved):

PR #766, opened 2023-12-04
document default payment term in paymentTerms.list

joaoinacio:


mrtus (dismissed):
💯

tiagobastos:
✔️ Manually checked

apiary.apib
```diff
@@ -4421,6 +4422,8 @@ Get a list of available payment terms.
             + (object)
                 + id: `c93ddb52-0af8-47d9-8551-441435be66a7` (string)
                 + Include PaymentTerm
+        + meta (object)
+            + `default`: `c93ddb52-0af8-47d9-8551-441435be66a7` (string)
```
mrtus, line 4426: Let's make sure we align this property structure with all future ones we'd do, e.g. dealPipelines cc @gdegirmenci

Jdww (dismissed):

mrtus (approved):

Jdww (approved):

apiary.apib
mrtus, line 4426: I've actually been having some doubts for the past couple of days about the naming, should we include what the default is about?

```
default_payment_term: `c93ddb52-0af8-47d9-8551-441435be66a7`
```


What if we introduce more "default" a-like values?

mrtus: Or apply more structure through
```
default: {
  id: c93ddb52-0af8-47d9-8551-441435be66a7,
  type: paymentTerm
}
```

mrtus: Or
```
default_payment_term: {
  id: c93ddb52-0af8-47d9-8551-441435be66a7,
  type: paymentTerm
}
```

Miljar: I'm actually fine with the original. My reasoning: We're on a `.list` endpoint and will return a single list (maybe paginated) of items of a single type. So `default` is for me clearly the default item from that list.

tiagobastos (approved):

apiary.apib
mrtus: We decided on keeping as is.

PR #785, opened 2024-01-19
Pagination on `projects-v2/tasks.list`

adognameddog:


lorgan3 (approved):

src/12-nextgen-projects/tasks.apib
```diff
@@ -12,6 +12,7 @@ Lists all the tasks that match the optional filters provided.
     + Attributes (object)
         + filter (object)
             + ids: `46156648-87c6-478d-8aa7-1dc3a00dacab`, `46156648-87c6-478d-8aa7-1dc3a00daca4` (array[string])
+        + `page` (Page, optional)
```
uhajzeraj, line 15: PP: By default, a parameter is optional, so the `optional` keyword could be dropped.

apiary.apib
```diff
@@ -6928,6 +6928,7 @@ Lists all the tasks that match the optional filters provided.
     + Attributes (object)
         + filter (object)
             + ids: `46156648-87c6-478d-8aa7-1dc3a00dacab`, `46156648-87c6-478d-8aa7-1dc3a00daca4` (array[string])
+        + `page` (Page, optional)
 
 + Response 200 (application/json)
```
uhajzeraj, line 6933: Q: Should we also include the pagination info in the `response.meta` ?

uhajzeraj (approved):

apiary.apib
lorgan3: I think we only add that when necessary because it has a performance cost

driesd (approved):

PR #799, opened 2024-02-13
Fix expiration time of access tokens

Klukies:
Hey 👋

I'm currently working an custom Teamleader integration for one of our clients and noticed a small bug in the authentication documentation.

As per the official [OAuth 2.0 RFC](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2) documentation, the `expires_in` denotes the the lifetime in seconds of the access token. As the access_token returned by the Teamleader when obtaining an access token is 3600 seconds, this expiration time of the access token is one hour, rather than 10 minutes.

### DIFF
--- a/src/authentication.apib
+++ b/src/authentication.apib
@@ -39,7 +39,7 @@ If the user denies your integration, he will be redirected to the `redirect_uri`
 
 ### Obtaining access tokens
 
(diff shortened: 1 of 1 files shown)


mrtus:
Hi @Klukies

Thank you for contributing!

You are correct when you say that our access tokens have a TTL of 1 hour.

I think the initial documentation is correct, however. The expiry time of 10 minutes seems to be reflected onto the expiry time of the authorization code and not that of the access token.
An access token is given in exchange for the shorter-lived one-time use authorization code.

Does that clarify your concern @Klukies?

Klukies:
Aha, yep that does indeed make sense, thanks for the clarification.
I must've overlooked that detail 😅, I'll close this PR then!

PR #831, opened 2024-07-30
custom fields on list endpoints docs

gehadMohamed:
API doc for custom fields in list endpoints for:
- deals.list
- companies.list
- projects-v2/projects.list

### DIFF
--- a/apiary.apib
+++ b/apiary.apib
@@ -1986,7 +1986,7 @@ Get a list of contacts.
                 + (object)
                     + field: `name`
                     + order: `asc`
-        + includes: `custom_fields` (string, optional) - when used, the response will include `custom_fields`
+        + includes: `custom_fields` (string, optional) - Comma-separated list of optional includes
 
 + Response 200 (application/json)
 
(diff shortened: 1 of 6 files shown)


StephenBeirlaen (dismissed):

eniskraasniqi1 (dismissed):

tiagobastos (dismissed):

gdegirmenci (dismissed):

pidera (dismissed):

adognameddog (dismissed):

adognameddog (commented):
Screenshot 2024-08-05 at 09 53 03

I don't feel like it's very clear this way. This sentence explains that you can provide any string, and the response will include custom fields. I realise that, because we use a concatenated string here, we can't use `enum` with `Members`.

My suggestion would be:

> includes: `custom_fields` (string, optional) - Comma-separated string of optional includes.

Also, explanations should be capitalised, so `When...`.

StephenBeirlaen (approved):

joaoinacio (approved):

joaoinacio:
✅

eniskraasniqi1 (approved):

PR #833, opened 2024-08-05
Document `/accounts/projects-v2-status`

adognameddog:


apiary.apib
```diff
@@ -9021,6 +9021,23 @@ Fetch cloudPlatform url for type and id
         + data (object)
             + url: `https://teamleader.cloud/eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.yUmR5yNZ45P_jHDbjAzuk4kRA8YNoM9ckSZOZpMIJmU/` (string)
 
+## Accounts [/accounts]
+
+### accounts.projects-v2-status [POST /accounts.projects-v2-status]
```
JordyDeGroeve, line 9026: Q: is the `v2` needed in the path? It makes more sense to make this general, no?
```suggestion
### accounts.projects-status [POST /accounts.projects-status]
```

JordyDeGroeve (approved):

apiary.apib
adognameddog: We discussed this, and the conclusion was that it can be construed as the status of a project or projects, where this means what the status is of the move to projects v2.

But a very sensible comment 👌🏻

lorgan3 (approved):

StijnVrolijk (approved):

PR #838, opened 2024-08-09
add note about partial update for custom fields

gehadMohamed:
Add note about partial update for custom fields under collections section in introduction.apib

### DIFF
--- a/src/introduction.apib
+++ b/src/introduction.apib
@@ -191,3 +191,9 @@ POST https://api.focus.teamleader.eu/contacts.update
   ]
 }
 ```
+
+> **Note:**
+>
+> + To ensure you get the most out of the API, we have enhanced the update endpoints to support partial updates for custom fields. This allows you to update only the specific fields you want to change, while retaining the values of the other fields.
+>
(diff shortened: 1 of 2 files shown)


src/introduction.apib
```diff
@@ -191,3 +191,9 @@ POST https://api.focus.teamleader.eu/contacts.update
   ]
 }
 ```
+
+> **Note:**
+>
+> + There is an exception for custom field collections where partial updates are possible, by including the parameter `custom_fields_update_strategy` with the value `partial` in the request body.
+>
+> + This will allow updating only values of specific custom fields and leave the remaining ones unchanged.
```
BrechtBonte: PP:
```suggestion
> + This will allow updating only the values of specific custom fields and leaving the remaining ones unchanged.
```

BrechtBonte (dismissed):

JordyDeGroeve (dismissed):

JordyDeGroeve (approved):

joaoinacio (approved):

StephenBeirlaen (approved):

eniskraasniqi1 (approved):

BrechtBonte (approved):

PR #854, opened 2024-11-06
subscriptions.info/create/update: support direct_debit payment method in invoice_generation

joaoinacio:


gehadMohamed (dismissed):
PP: It could be nullable in both create and update requests

It's a little meaningless as it's also optional, but at least we will have the exact contract for both

StephenBeirlaen (dismissed):

gehadMohamed:
✅ checked

gehadMohamed (approved):

BrechtBonte (approved):

adognameddog (approved):

StijnVrolijk (approved):

tiagobastos (approved):

PR #856, opened 2024-12-04
`project` type in `notes.list` and "Initial values" in `projects.create`

adognameddog:


mrtus (dismissed):

StephenBeirlaen (dismissed):

apiary.apib
```diff
@@ -9690,15 +9705,16 @@ Fetch which version of Projects the account is using.
...
 + `company`
++ `contact`
++ `creditNote`
 + `deal`
 + `invoice`
-+ `creditNote`
-+ `subscription`
++ `nextgenProject`
 + `product`
++ `project`
```
lorgan3: [notes.create](https://github.com/teamleadercrm/api/blob/master/src/01-general/notes.apib#L35) does not allow passing a project though.
These types should be split up since it doesn't make sense to still add support old projects in this endpoint

mrtus (approved):

StijnVrolijk (approved):

StijnVrolijk:
✅ It renders correctly

StephenBeirlaen (approved):

lorgan3 (approved):

